PR #2, opened 2016-04-24
Don't pass arbitrary strings as Errorf() format strings.

enisoc:
Sometimes they happen to contain format specifiers, which results in
things like "%!v(MISSING)".


davecheney:
Thanks for to fixing that, fmt.Errof was me being lazy

PR #7, opened 2016-04-24
Add Wrapf

### LINKED ISSUE #6: Add Wrapf(cause error, format string, args ...interface{}) error
Instead of this

func Wrap(cause error, message string) error

Have these two 

func Annot(cause error, a ...interface{}) error
func Annotf(cause error, format string, a ...interface{}) error

In my opinion, Wrap() doesn't convey the purpose of the function as well as Annot() does. Finally, supporting variadics keeps things consistent with everything else.

umairidris:
Fixes #6

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -113,6 +113,20 @@ func Wrap(cause error, message string) error {
 	}
 }
 
+// Wrapf returns an error annotating the cause with the format specifier.
+// If cause is nil, Wrapf returns nil.
+func Wrapf(cause error, format string, args ...interface{}) error {
+	if cause == nil {
+		return nil
+	}
+	pc, _, _, _ := runtime.Caller(1)
+	return &e{
+		cause:   cause,
+		message: fmt.Sprintf(format, args...),
+		loc:     loc(pc),
+	}
+}
+
 type causer interface {
 	Cause() error
 }
--- a/errors_test.go
+++ b/errors_test.go
@@ -130,6 +130,9 @@ func TestFprint(t *testing.T) {
 	}, {
 		err:  Wrap(Wrap(x, "message"), "another message"),
 		want: "github.com/pkg/errors/errors_test.go:131: another message\ngithub.com/pkg/errors/errors_test.go:131: message\ngithub.com/pkg/errors/errors_test.go:106: error\n",
+	}, {
+		err:  Wrapf(x, "message"),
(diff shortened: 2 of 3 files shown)


errors_test.go
```diff
@@ -53,6 +53,31 @@ func TestWrap(t *testing.T) {
 	}
 }
 
+func TestWrapfNil(t *testing.T) {
```
davecheney: I'd you add these tests below the next block you won't have to adjust the line numbers below

errors.go
```diff
@@ -113,6 +113,12 @@ func Wrap(cause error, message string) error {
 	}
 }
 
+// Wrapf returns an error annotating the cause with the format specifier.
+// If cause is nil, Wrapf returns nil.
+func Wrapf(cause error, format string, args ...interface{}) error {
+	return Wrap(cause, fmt.Sprintf(format, args...))
```
davecheney: Please add a test to ensure that the correct line number is being recorded. I think this code will show that Wrapf is the owner, not it's caller

davecheney:
Thank you for this PR . please also add an example test for Wrapf

umairidris:
My apologies for the glaring bug. I am not sure what the ideal way to fix it, but for the time being I have created a private function wrap which both Wrap and Wrapf can uniformly call.

I have also moved the test cases below the Fprint tests so the line numbers no longer need to be changed.

A test case checking the Wrapf line number as well as an example have been added.

Thanks!
Umair

errors.go
```diff
@@ -99,20 +99,30 @@ func (e *e) Cause() error {
 	return e.cause
 }
 
-// Wrap returns an error annotating the cause with message.
-// If cause is nil, Wrap returns nil.
-func Wrap(cause error, message string) error {
+func wrap(cause error, message string) error {
```
davecheney: This change is bigger than it needs to be. I suggest copying Wrap to Wrapf and changing the call to errors.New with a fmt.Errorf

umairidris: Done.

davecheney: Sorry, that's not what I meant, just copy Wrap to Wrapf and use fmt.Error
in place of errors.New. delete wrap

umairidris: Sorry for this getting dragged on a bit longer than it needed to. Copied Wrap to Wrapf and squashed into one commit.

davecheney:
LGTM. Thank you.

Issue #8, opened 2016-04-25
Incorrect file returned by Location when vendored or in a different part of a multipart GOPATH.


ChrisHines:
To reproduce:

```
$ go get github.com/ChrisHines/pkg-errors-test
$ pkg-errors-test
ors-test.go:8: where am I?
```

To reproduce the multipart GOPATH scenario copy the pkg-errors-test repo to a different part of a mulitpart GOPATH from github.com/pkg/errors and delete the vendor folder. In that case you might see something like:

```
$ go run pkg-errors-test.go
hub.com/ChrisHines/pkg-errors-test/pkg-errors-test.go:8: where am I?
```

The exact output will depend on the path length delta between the two parts of the GOPATH.

I believe this can be fixed using a technique like what I have done here: https://github.com/go-stack/stack/blob/master/stack.go#L96.

I will work on a PR for this issue.

davecheney:
Thank you. Please do.

PR #9, opened 2016-04-25
Infer compile time GOPATH from fn.Name().

ChrisHines:
Fix for #8.

All existing tests still pass, and I have tested this against both scenarios mentioned in issue #8 but I cannot think of a convenient way to test the issue directly from within this repository. We could potentially do a bunch of file manipulations to generate a package structure in a temporary folder and run the test as a child process, but that might be going too far. Thoughts?

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -91,15 +91,15 @@ func (l loc) Location() (string, int) {
 	// leading separator.
 	const sep = "/"
 	goal := strings.Count(fn.Name(), sep) + 2
-	pathCnt := 0
 	i := len(file)
-	for pathCnt < goal {
+	for n := 0; n < goal; n++ {
 		i = strings.LastIndex(file[:i], sep)
 		if i == -1 {
+			// not enough separators found, set i so that the slice expression
+			// below leaves file unmodified
 			i = -len(sep)
 			break
 		}
-		pathCnt++
 	}
 	// get back to 0 or trim the leading seperator
 	file = file[i+len(sep):]
--- a/errors.go
+++ b/errors.go
@@ -65,11 +65,45 @@ func (l loc) Location() (string, int) {
 		return "unknown", 0
 	}
 
-	_, prefix, _, _ := runtime.Caller(0)
 	file, line := fn.FileLine(pc)
-	if i := strings.LastIndex(prefix, "github.com/pkg/errors"); i > 0 {
-		file = file[i:]
+
+	// Here we want to get the source file path relative to the compile time
+	// GOPATH. As of Go 1.6.x there is no direct way to know the compiled
+	// GOPATH at runtime, but we can infer the number of path segments in the
+	// GOPATH. We note that fn.Name() returns the function name qualified by
+	// the import path, which does not include the GOPATH. Thus we can trim
+	// segments from the beginning of the file path until the number of path
+	// separators remaining is one more than the number of path separators in
+	// the function name. For example, given:
+	//
+	//    GOPATH     /home/user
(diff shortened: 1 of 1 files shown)


ChrisHines:
One side effect of this change is that errors produced in package main will not have their full package path used, but only the last segment of the package path. This is because `runtime.Func.Name` does not include the full import path as a prefix for functions in package main. For example, with this PR vendored, the test repo I posted outputs:

```
pkg-errors-test/pkg-errors-test.go:8: where am I?
```

It would be nice if it produced the following instead.

```
github.com/ChrisHines/pkg-errors-test/pkg-errors-test.go:8: where am I?
```

This result is mildly unfortunate. Errors produced by non-main packages should work as desired. I think it is a net improvement, especially when `github.com/pkg/errors` is vendored.

davecheney:
Thanks Chris, I'm not too worried about this restriction, as it works
towards encouraging people to keep their main packages small.

errors.go
```diff
@@ -65,11 +65,45 @@ func (l loc) Location() (string, int) {
...
+	//    pkg/sub/file.go
+	//
+	// From this we can easily see that fn.Name() has one less path separator
+	// than our desired output. We count separators from the end of the file
+	// path until it finds two more than in the function name and then move
+	// one character forward to preserve the initial path segment without a
+	// leading separator.
+	const sep = "/"
+	goal := strings.Count(fn.Name(), sep) + 2
+	pathCnt := 0
```
davecheney: Would you move this into the for loop, 

for pathCnt := 0; pathCnt < goal; pathCnt++

And while your there name the loop variable n, not pathCnt

errors.go
```diff
@@ -65,11 +65,45 @@ func (l loc) Location() (string, int) {
...
+	i := len(file)
+	for pathCnt < goal {
+		i = strings.LastIndex(file[:i], sep)
+		if i == -1 {
+			i = -len(sep)
```
davecheney: I -= len(sep)

ChrisHines:
I am not worried about it either. Just providing full disclosure. This logic has been in github.com/go-stack/stack and github.com/inconshreveable/log15 for a long time and I've never received any complaints.

errors.go
```diff
@@ -65,11 +65,45 @@ func (l loc) Location() (string, int) {
...
+	// We want to produce:
+	//
+	//    pkg/sub/file.go
+	//
+	// From this we can easily see that fn.Name() has one less path separator
+	// than our desired output. We count separators from the end of the file
+	// path until it finds two more than in the function name and then move
+	// one character forward to preserve the initial path segment without a
+	// leading separator.
+	const sep = "/"
```
davecheney: I cannot test atm, but does this work on windows?

davecheney:
Thanks Chris, this looks pretty good with some very minor nit picking comments

errors.go
ChrisHines: Yes. Oddly package runtime uses "/" for these paths even on Windows. I wrote and tested this on Windows, but have used this logic also on Linux.

Working on fixing the nits. Not sure why that variable wasn't part of the for loop to begin with.

errors.go
ChrisHines: That doesn't do the same thing.

ChrisHines: I will add a comment to clarify the logic.

ChrisHines:
@davecheney PTAL

errors.go
davecheney: Thanks for checking.

davecheney:
LGTM. Thanks!

PR #12, opened 2016-04-26
make func (l location) Location() (string, int) simpler

Tang-RoseChild:


ChrisHines:
This must be tested with a package where directories in the package import path contain a period `.`.

ChrisHines:
Scanning the file path backwards is fundamental to correct behavior in order to avoid problems when the file path contains multiple occurrences of the package import string.

For example, the proposed code would produce the wrong output given the following conditions:

```
GOPATH     /home/user
file       /home/user/src/user/file.go
fn.Name()  user.Fn
```

In addition, the code in this PR would return `"unknown", 0` for all functions in most main packages. The exception being if `"main"` appeared somewhere in the file path, but it would still return misleading information in that case.

```
GOPATH     /home/user
file       /home/user/src/user/my-app.go
fn.Name()  main.main
```

Tang-RoseChild:
i got it. Thanks

PR #14, opened 2016-04-26
Add Errorf

### LINKED ISSUE #13: Add errors.Errorf method
Hi, would it make sense to have a Newf method to avoid having to write this?

     errors.New(fmt.Sprintf("something %v", x))

If you think it does I can make a PR.

Thank you

scorredoira:
Fixes #13

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -130,6 +130,19 @@ func (c cause) Error() string   { return c.Message() + ": " + c.Cause().Error() 
 func (c cause) Cause() error    { return c.cause }
 func (c cause) Message() string { return c.message }
 
+// Errorf formats according to a format specifier and returns the string
+// as a value that satisfies error.
+func Errorf(format string, args ...interface{}) error {
+	pc, _, _, _ := runtime.Caller(1)
+	return struct {
+		error
+		location
+	}{
+		fmt.Errorf(format, args...),
+		location(pc),
+	}
+}
+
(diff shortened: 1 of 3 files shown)


davecheney:
Thank you. Before I can merge this, please add tests and examples.

errors.go
```diff
@@ -130,6 +130,18 @@ func (c cause) Error() string   { return c.Message() + ": " + c.Cause().Error()
 func (c cause) Cause() error    { return c.cause }
 func (c cause) Message() string { return c.message }
 
+// Errorf returns an error that formats as the given text.
```
davecheney: Errorf formats according to a format specifier and returns the string as a value that satisfies error.

paulstuart:
I know this is late to the dance, but why not have optional arguments to New and treat it as printf?

davecheney:
@paulstuart probably for the same reason as https://github.com/pkg/errors/issues/6#issuecomment-214054015

Please open a new issue to discuss this, I'm about to merge this PR and this discussion may be lost.

paulstuart:
@davecheney the existing solution is fine. It just seemed like the idea was a clean and minimal enhancement. My itch is scratched. Keep doing the good you do.

davecheney:
Hmm, so New() could be redefined as New(string, ...string) but I'm not sure how useful that would be, if you want more complicated formatting, Errorf would probably be the right tool for the job.

PR #19, opened 2016-05-03
Add WithLocation

faxal:
Currently it is not possible to add location information without adding more context information. This adds a function `WithLocation` that just adds location information to an error. 

Sometimes the only needed context is location.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -109,16 +109,25 @@ func (l location) Location() (string, int) {
 	return file, line
 }
 
-// New returns an error that formats as the given text.
-func New(text string) error {
-	pc, _, _, _ := runtime.Caller(1)
+func newErr(err error) error {
+	pc, _, _, _ := runtime.Caller(2)
 	return struct {
 		error
 		location
 	}{
-		errors.New(text),
+		err,
 		location(pc),
 	}
+}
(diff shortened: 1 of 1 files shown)


errors.go
```diff
@@ -109,18 +109,27 @@ func (l location) Location() (string, int) {
...
 		location(pc),
 	}
 }
 
+// New returns an error that formats as the given text.
+func New(text string) error {
+	return newErr(errors.New(text))
+}
+
+// WithLocation is like New but wraps the location for an existing error
```
seh: Punctuate this sentence.

faxal: Done.

davecheney:
Please open an issue to discuss adding this feature before slinging code. Thanks

davecheney:
Not LGTM, sorry. 

You can already do this with `errors.Wrap(err, "")`.

davecheney:
One possible way to do this would be to make Wrap varadic, so it takes an error and a variadic number of string arguments. I'm not sure if this pattern of wanting to wrap to record the location is common enough to warrant this.

Again, please raise an issue to discuss this before we commit to writing code.

Thanks

Dave

PR #30, opened 2016-05-24
Make errors.stack comparable

### LINKED ISSUE #29: panic: values with embedded slices uncomparable errors
Please see the stack trace here: https://travis-ci.org/opencontainers/image-spec/builds/132403826, described by opencontainers/image-spec#84.

Looks like pkg/errors#27 introduced or exposed a new error type that uses a slice and thus cannot be compared without introducing a panic. We picked this change up through an update to blackfriday, but this could happen anywhere.

While I understand that this library is based on concepts of no longer using sentinel values, this is already a common practice in existing code. There is also limited ways for a caller to avoid the panic.

I suggest that errors with a slice value use a pointer to the error value to avoid this trap in the future. For example, instead of building a `stack{}`, use `&stack{}`. This will ensure that existing code can make error comparisons while new code can begin to adopt this great package.

davecheney:
Fixed #29

errors.New, etc values are not expected to be compared by value but
the change in errors#27 made them incomparable. Assert that various
kinds of errors have a functional equality operator, even if the
result of that equality is always false.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -64,17 +64,17 @@ import (
 // stack represents a stack of programm counters.
 type stack []uintptr
 
-func (s stack) Stack() []uintptr { return s }
+func (s *stack) Stack() []uintptr { return *s }
 
-func (s stack) Location() (string, int) {
-	return location(s[0] - 1)
+func (s *stack) Location() (string, int) {
+	return location((*s)[0] - 1)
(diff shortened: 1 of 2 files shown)


s-urbaniak:
That was fast, thanks!

PR #37, opened 2016-06-06
Introduce Stacktrace and Frame

davecheney:
This PR is a continuation of a series aimed at exposing the stack trace
information embedded in each error value. The secondary effect is to
deprecated the `Fprintf` helper.

Taking cues from from @chrishines' `stack` package this PR introduces a
new interface `Stacktrace() []Frame` and a `Frame` type, similar in
function to the `runtime.Frame` type (although lacking its iterator
type). Each `Frame` implemnts `fmt.Formatter` allowing it to print
itself.

The older `Location` interface is still supported but also deprecated.

### DIFF
--- a/stack.go
+++ b/stack.go
@@ -47,6 +47,7 @@ func (f Frame) line() int {
 // Format accepts flags that alter the printing of some verbs, as follows:
 //
 //    %+s   path of source file relative to the compile time GOPATH
+//    %+v   equivalent to %+s:%d
 func (f Frame) Format(s fmt.State, verb rune) {
 	switch verb {
 	case 's':
@@ -54,7 +55,12 @@ func (f Frame) Format(s fmt.State, verb rune) {
 		case s.Flag('+'):
 			pc := f.pc()
 			fn := runtime.FuncForPC(pc)
-			io.WriteString(s, trimGOPATH(fn, pc))
+			if fn == nil {
+				io.WriteString(s, "unknown")
+			} else {
+				file, _ := fn.FileLine(pc)
+				io.WriteString(s, trimGOPATH(fn.Name(), file))
+			}
 		default:
 			io.WriteString(s, path.Base(f.file()))
 		}
@@ -62,7 +68,9 @@ func (f Frame) Format(s fmt.State, verb rune) {
 		fmt.Fprintf(s, "%d", f.line())
 	case 'n':
 		name := runtime.FuncForPC(f.pc()).Name()
-		i := strings.LastIndex(name, ".")
+		i := strings.LastIndex(name, "/")
+		name = name[i+1:]
+		i = strings.Index(name, ".")
 		io.WriteString(s, name[i+1:])
 	case 'v':
 		f.Format(s, 's')
@@ -99,7 +107,7 @@ func callers() *stack {
 	return &st
 }
 
-func trimGOPATH(fn *runtime.Func, pc uintptr) string {
+func trimGOPATH(name, file string) string {
 	// Here we want to get the source file path relative to the compile time
 	// GOPATH. As of Go 1.6.x there is no direct way to know the compiled
 	// GOPATH at runtime, but we can infer the number of path segments in the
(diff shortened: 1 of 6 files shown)


stack.go
```diff
@@ -0,0 +1,141 @@
...
+			fn := runtime.FuncForPC(pc)
+			io.WriteString(s, trimGOPATH(fn, pc))
+		default:
+			io.WriteString(s, path.Base(f.file()))
+		}
+	case 'd':
+		fmt.Fprintf(s, "%d", f.line())
+	case 'n':
+		name := runtime.FuncForPC(f.pc()).Name()
+		i := strings.LastIndex(name, ".")
```
ChrisHines: Note that for a method `name` will look like *package-path*.*type-name*.*method-name*. Using `LastIndex(name, ".")` excludes the type-name, which would be ambiguous when many methods have the same name. That is why the stack package finds the first `.` after the last `/`.

stack.go
```diff
@@ -0,0 +1,141 @@
...
+// Format formats the frame according to the fmt.Formatter interface.
+//
+//    %s    source file
+//    %d    source line
+//    %n    function name
+//    %v    equivalent to %s:%d
+//
+// Format accepts flags that alter the printing of some verbs, as follows:
+//
+//    %+s   path of source file relative to the compile time GOPATH
```
ChrisHines: %+v is also supported because `case 'v'` recursively calls `f.Format(s, 's')`.

stack.go
```diff
@@ -0,0 +1,141 @@
...
+func (f Frame) Format(s fmt.State, verb rune) {
+	switch verb {
+	case 's':
+		switch {
+		case s.Flag('+'):
+			pc := f.pc()
+			fn := runtime.FuncForPC(pc)
```
ChrisHines: `runtime.FuncFoPC` docs say it can return `nil`.

stack.go
davecheney: Thanks. I cheated there. I'll try to write a test case and fix this.

stack.go
davecheney: I didn't want to add it til I had written a test case for it (I try to TDD where I can). As you say, it'll probably work because of the way I wrote `%v`

stack.go
davecheney: yes, that is true, I'm leaning on the `recover` inside the `fmt` package to catch me when I fall. 

Can you think of a way we could write a test case for this ? We might have to extract the body of `Format` into a function so we can pass in an arbitrary `pc`.

ChrisHines: Add this test to `TestFrameFormat`:

```go
	}, {
		Frame(0),
		"%+s",
		"unknown",
	}, {
```

It will fail. Fix it for TDD points.

stack_test.go
```diff
@@ -0,0 +1,128 @@
+package errors
+
+import (
+	"fmt"
+	"runtime"
+	"testing"
+)
+
+var line, _, _, _ = runtime.Caller(0)
```
ChrisHines: The variable name `line` is misleading because it actually holds the `pc` value at this point. Consider the name `initpc`.

stack_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestFrameLine(t *testing.T) {
+	var tests = []struct {
+		Frame
+		want int
+	}{{
+		Frame(line),
+		9,
+	}, {
+		func() Frame {
+			var line, _, _, _ = runtime.Caller(0)
```
ChrisHines: s/line/pc (and more like it below).

stack_test.go
davecheney: Done

stack_test.go
davecheney: Done, doh

stack.go
davecheney: Done

stack.go
davecheney: Added test and docs.

PR #40, opened 2016-06-08
Use fmt.Formatter throughout

davecheney:
Replace `Fprintf` with `fmt.Formatter` logic on various types. This
effectively guts `Fprint` to be just a recursive call down the `err.Cause()`
chain. The next step will be to move this recursive logic into
`wrapper.Format`.

This change necessitates adding types for the error implementations returned from
`New` and `Errorf`, and `Wrap` and `Wrapf`, respectively. The name of the latter
type is acceptable, the former is not and alternative suggestions are
encouraged.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -59,12 +59,28 @@ import (
 	"io"
 )
 
+// _error is an error implementation returned by New and Errorf
+// that implements its own fmt.Formatter.
+type _error struct {
+	error
+	*stack
+}
+
+func (e _error) Format(s fmt.State, verb rune) {
+	switch verb {
+	case 'v':
+		if s.Flag('+') {
+			fmt.Fprintf(s, "%+v: ", e.Stacktrace()[0])
+		}
+		fallthrough
+	case 's':
+		io.WriteString(s, e.Error())
+	}
+}
+
 // New returns an error that formats as the given text.
 func New(text string) error {
-	return struct {
-		error
-		*stack
-	}{
+	return _error{
 		errors.New(text),
 		callers(),
 	}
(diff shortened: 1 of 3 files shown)


errors.go
```diff
@@ -87,19 +101,26 @@ type cause struct {
...
+
+func (w wrapper) Format(s fmt.State, verb rune) {
 	switch verb {
 	case 'v':
 		if s.Flag('+') {
-			io.WriteString(s, c.msg)
+			fmt.Fprintf(s, "%+v: %s", w.Stacktrace()[0], w.cause.msg)
 			return
 		}
 		fallthrough
```
ChrisHines: This logic seems strangely different than the code for the `+` flag in `_error.Format`. It seems to me that it should look like this:

```go
		if s.Flag('+') {
			fmt.Fprintf(s, "%+v: ", w.Stacktrace()[0])
		}
		fallthrough
```

The tests cover this code, but only indirectly via testing `errors.Fprint`. A test that calls `fmt.Printf("%+v", wrappedErr)` would help clarify the proper behavior here.

davecheney: This might be being distorted by the front tests. My aim once this is
committed it is delete fprint and then recheck the coverage of all helpers.

ChrisHines:
Alternate name ideas for the type returned from `New` and `Errorf`: `base`, `message`, `errorString`, `msgError`.

davecheney:
I probably shouldn't tell you that _error and wrapper were originally
called e and w

ChrisHines:
The parameter names of `text` and `message` for the functions `New` and `Wrap` respectively are not consistent, despite the similar semantics.

ChrisHines:
LGTM with the proviso regarding `%+v` formatting in `wrapped.Format`.

davecheney:
I'll fix the parameters names in a separate pr, point taken.

davecheney:
@ChrisHines have a look at the fmt.Formatter tests in 15a7541

PR #44, opened 2016-06-09
Reimplement Fprint in terms of fmt.Fprintf

davecheney:
This PR follows on from #40 completely implementing Fprint in terms of
fmt.Fprintf. This will be the final PR before Fprint is removed.


ChrisHines:
LGTM.

PR #54, opened 2016-06-14
wrapper should serialize nicely to JSON

joshuarubin:
without this patch, `wrapper` serializes to `JSON` similar to this:

```
{"error":{"stack":[534500,534894,525574,10756,389870,1776797,1771873,693326,682837,10507,202432,399841]}}
```

with this patch, `wrapper` serializes to `JSON` like this:

```
{"error":"error creating file: open dir: is a directory"}
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -140,6 +140,10 @@ type wrapper struct {
 	*stack
 }
 
+func (w wrapper) MarshalJSON() ([]byte, error) {
+	return []byte(`"` + w.Error() + `"`), nil
+}
+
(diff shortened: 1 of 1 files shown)


davecheney:
@joshuarubin thank you for your proposal. The policy if this repository is to discuss design before coding. Please open an issue.

To be clear, I'm not opposed to adding json marshalling to error types, but there needs to be _one_ format that all consumers of this JSON data (I am not one of them) can agree on.

PR #55, opened 2016-06-16
Make example interfaces unnamed

davecheney:
To avoid bike shedding over the name, and make it clear that these
interface defintions are _not_ provided by this package.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -28,7 +28,7 @@
 // to reverse the operation of errors.Wrap to retrieve the original error
 // for inspection. Any error value which implements this interface
 //
-//     type Causer interface {
+//     interface {
 //             Cause() error
 //     }
 //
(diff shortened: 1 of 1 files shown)


seh:
I am happy to see this change. I've found this technique of _showing_ notional interfaces that are not actually defined by the name to be confusing. Clients of the library can't actually refer to that name, nor even find it in the source code apart from in a comment. Better is to omit the name as you've done here.

errors.go
```diff
@@ -59,7 +59,7 @@
 // New, Errorf, Wrap, and Wrapf record a stack trace at the point they are
 // invoked. This information can be retrieved with the following interface.
 //
-//     type StackTrace interface {
+//     interface {
```
seh: What about the example cast that mentions the `StackTrace` type [a few lines down](https://github.com/pkg/errors/blob/45fa7bc37676ec7003f63dad4dda3955012806d8/errors.go#L74)? Should that change as well?

davecheney: Well, that does make things complicated.

davecheney:
Abandoned.

@ser @kr if you want to send a PR to change the suffix please do so, making the type private for good measure. Please also change the declaration in example_test.go

PR #58, opened 2016-06-22
Fix %q format for wrapped errors

greensnark:
Not handling the %q format causes logrus to log empty error strings for
wrapped errors where x.Error() contains spaces or other quoteworthy
characters.

For reference, the logrus formatter does this:
https://github.com/Sirupsen/logrus/blob/master/text_formatter.go#L154


davecheney:
LGTM. Thanks

PR #60, opened 2016-06-25
Capitalise first letter of trace.

jongillham:
For consistency reasons I thought I would capitalise the first letter of the remaining 'traces' seeing as `Stacktrace` has now been changed to `StackTrace`.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -59,7 +59,7 @@
 // New, Errorf, Wrap, and Wrapf record a stack trace at the point they are
 // invoked. This information can be retrieved with the following interface.
 //
-//     type stacktracer interface {
+//     type stackTracer interface {
 //             StackTrace() errors.StackTrace
 //     }
 //
@@ -67,11 +67,11 @@
 //
 //     type StackTrace []Frame
 //
-// The Frame type represents a call site in the stacktrace. Frame supports
+// The Frame type represents a call site in the stack trace. Frame supports
 // the fmt.Formatter interface that can be used for printing information about
-// the stacktrace of this error. For example:
+// the stack trace of this error. For example:
 //
-//     if err, ok := err.(stacktracer); ok {
+//     if err, ok := err.(stackTracer); ok {
 //             for _, f := range err.StackTrace() {
 //                     fmt.Printf("%+s:%d", f)
 //             }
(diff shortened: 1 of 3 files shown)


errors.go
```diff
@@ -59,19 +59,19 @@
 // New, Errorf, Wrap, and Wrapf record a stack trace at the point they are
 // invoked. This information can be retrieved with the following interface.
 //
-//     type stacktracer interface {
```
davecheney: No thank you

errors.go
```diff
@@ -59,19 +59,19 @@
...
 //
 //     type StackTrace []Frame
 //
-// The Frame type represents a call site in the stacktrace. Frame supports
+// The Frame type represents a call site in the stack trace. Frame supports
 // the fmt.Formatter interface that can be used for printing information about
-// the stacktrace of this error. For example:
+// the stack trace of this error. For example:
 //
-//     if err, ok := err.(stacktracer); ok {
```
davecheney: No thank you

example_test.go
```diff
@@ -119,14 +119,14 @@ func ExampleErrorf_extended() {
 	//         /home/dfc/go/src/runtime/asm_amd64.s:2059
 }
 
-func Example_stacktrace() {
-	type stacktracer interface {
+func Example_stackTrace() {
```
davecheney: No thanks you

davecheney:
I'm sorry. I'm having a hard time getting enthused about renaming private symbols.

jongillham:
Hi @davecheney, as a learning point for me, why? I assumed changing the symbols by using camel case to separate the words would make the code base more consistent, irrespective of whether they are public or private. Maybe there are code conventions or repercussions I don't know about? Maybe it's personal style?

davecheney:
Im sorry, on reflection as you've done the work it seems silly to not merge this.

For reference, my hope was that we could get away with giving the interface type _any_ name as the name is not the important bit, the `StackTrace` method is the important bit.

jongillham:
@davecheney I see. I suppose the following pattern could be used in the examples to make your point that the interface type name is not important:

    If err, ok := err.(interface {
    }); ok {
        // Use the stack trace.
    }

It doesn't look to pretty though which might be a good reason not to use it.

davecheney:
Yeah. Gofmt doesn't make that look pretty at all.

PR #65, opened 2016-06-28
document goimports issue with New, fix #62

### LINKED ISSUE #62: goimports miss
For many files it'll be enough to use just `errors.New()`, which result in goimports incorrectly import "errors" instead of "github.com/pkg/errors". This mistake will be hard to spot manually while code review and it'll show itself only when one of these errors will be output into log without stack trace.

Not sure is you can do anything with this - maybe just note this subtle trap in package documentation?

powerman:
Fixes #62

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -109,6 +109,10 @@ func (e _error) Format(s fmt.State, verb rune) {
 }
 
 // New returns an error with the supplied message.
+//
+// If New is only used function from this package in file processed
+// by goimports then you'll have to manually import this package,
+// otherwise stdlib's errors.New() will be called instead.
 func New(message string) error {
 	return _error{
 		message,


umairidris:
Perhaps more succinct:

Note: If New is the only function used from this package, goimports will import the standard library's errors package instead.

Though having to document another library doing something by accident may open a can of worms. For example, should all libraries that can experience this document it as well? Or, if another library like goimports has a similar problem, should we document that library as well? 

Perhaps documentation like this belongs in the goimports library not here.

davecheney:
I feel the same way. I recognise the problem, but feel disgruntled that I have to document this in my program because someone else's program gets it wrong.

powerman:
I 100% agree - in general. But in this one concrete case with `errors.New` conflict with stdlib and so much wide used tool as goimports - I think it may worth to warn users about this issue.

errors.go
```diff
@@ -109,6 +109,10 @@ func (e _error) Format(s fmt.State, verb rune) {
 }
 
 // New returns an error with the supplied message.
+//
+// If New is only used function from this package in file processed
```
davecheney: // Be aware that tools like goimports may mistake this function for the standard
// libraries New function.

ches: *library's*

davecheney: please send a pr

davecheney:
This PR is dead.

powerman:
It's not dead, but I'm a bit busy with real work and I'm not a native English speaker, and you wanna fine-tune English phrases, so maybe it's better if you'll do it yourself.

frou:
> I recognise the problem, but feel disgruntled that I have to document this in my program because someone else's program gets it wrong.

How is goimports "getting it wrong"? If all someone has written is `errors.New(...)`, then goimports would have to be a mindreader to know that they actually didn't want the standard library.

PR #69, opened 2016-07-11
Documentation improvements

### LINKED ISSUE #66: What is the reason behind stacktrace being private?
According to errors pkg documentation to retrieve the stack trace of an error or wrapper I should try to use interface:

```type stackTracer interface {
}```

The stackTracer interface is private. Does it mean, that you don't want it to be an API for errrors library and it can change its behaviour?

AlekSi:
Fixes #66.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Package errors provides simple error handling primitives.
 
+`go get github.com/pkg/errors`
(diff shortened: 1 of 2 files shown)


davecheney:
LGTM. Thanks.

PR #76, opened 2016-07-28
Destructure Wrap/Wrapf into component operations

davecheney:
This PR is a result of the discussions at GopherCon. The focus of the branch is to decompose Wrap (and Wrapf) into their component operations, namely adding a message, `errors.WithMessage` and adding a stack trace, `errors.WithStack` which were previously merged into one operation with Wrap.

This is accomplished by adding two top level functions, `errors.WithMessage` and `errors.WithStack` and rewriting Wrap and Wrap to work in terms of these operations although a small deviation is present to make sure the recorded stack trace is correct.

The motivation for this change was need to treat each of the following operations as distinct:

- Adding a context message to an existing error _without_ altering the stack trace.
- Adding a stack trace to an existing error _without_ the requirement adding and additional message.
- Retrieving the immediate cause of an error; popping one element of the error stack.

TODO:

- [x] add tests for WithMessage, WithStack 
- [x] add tests for various permutations of WithMessage, WithStack occurring directly rather than via Wrap/Wrapf.
- [x] ensure WithMessage properly delegates to its nearest parent Cause that supports StackTrace()

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -202,7 +202,7 @@ func Wrapf(err error, format string, args ...interface{}) error {
 }
 
 // WithMessage annotates err with a new message.
-// If err is nil, WithStack returns nil.
+// If err is nil, WithMessage returns nil.
 func WithMessage(err error, message string) error {
 	if err == nil {
 		return nil
--- a/errors.go
+++ b/errors.go
@@ -134,6 +134,18 @@ func (f *fundamental) Format(s fmt.State, verb rune) {
 	}
 }
 
+// WithStack annotates err with a stack trace at the point WithStack was called.
+// If err is nil, WithStack returns nil.
+func WithStack(err error) error {
+	if err == nil {
+		return nil
+	}
+	return &withStack{
+		err,
+		callers(),
+	}
+}
+
 type withStack struct {
 	error
 	*stack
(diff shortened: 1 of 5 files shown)


fabstu:
Hey!
I don't know about the discussions but I like the API additions, so I took a stab at adding some tests.

Testing mixed output was not possible with a simple regexp, so I wrote `parseBlocks` which splits up the output into message- and stacktrace strings quite nicely. Then I copied and adapted TestFormatWithMessage over to try the new parsing out and it seems to work. If you like this I can convert over the other format tests too.

https://github.com/fabstu/errors/commit/ef7cea437e4ececfb13e3364a5696dc2bf846ebc#diff-de087db9e6874e7d80f12c83ff178343R262

errors.go
```diff
@@ -193,6 +205,18 @@ func Wrapf(err error, format string, args ...interface{}) error {
 	}
 }
 
+// WithMessage annotates err with a new message.
+// If err is nil, WithStack returns nil.
```
cep21: s/WithStack/WithMessage/

davecheney: @cep21 please send a PR to this PR

nmiyake:
Added PR to this PR to address the comment issue.

Anything else I can help with to advance this PR? I can fix merge conflicts with master if you let me know whether you prefer merging master in or rebasing the PR on master and fixing that way. If there are any TODO items that could be tackled I could take a look as well -- would really like to see this functionality soon!

dragonsinth:
I know this may be a bit silly, but WithMessagef would be nice addition as well.  When converted "too many stack traces" code from Wrap -> WithMessage, converting Wrapf means an internal fmt.Sprintf().

davecheney:
https://github.com/pkg/errors/pull/118#issuecomment-309649529

PR #78, opened 2016-08-01
Modify TestTrimGOPATH to ensure tests pass while vendored

### LINKED ISSUE #77: TestTrimGOPATH fails when vendored
Due to using a simple string inequality test this function will fail if the errors package is located in a vendor subdirectory:

```
--- FAIL: TestTrimGOPATH (0.00s)
        stack_test.go:165: stack_test.go:9: want "github.com/pkg/errors/stack_test.go", got "example.org/test/vendor/github.com/pkg/errors/stack_test.go"
```

 A simple fix is to replace lines 163-166 with the same testFormatRegex function used elsewhere:
```
want := tt.want
if want != got {
	t.Errorf("%v: want %q, got %q", tt.Frame, want, got)
}
```

to

`testFormatRegexp(t, got, "%s", tt.want)`

This also requires changing the line numbers on the rest of the file, so a pull request is incoming shortly. This will also slightly change the error message returned when this test fails but I am not sure if that of any great consequence.

exp:
Fixes #77 

This should fix #77 .. I've tested it vendored and unvendored locally. I'm not sure what other tests you'd like to run.

### DIFF
--- a/stack_test.go
+++ b/stack_test.go
@@ -160,10 +160,7 @@ func TestTrimGOPATH(t *testing.T) {
 		fn := runtime.FuncForPC(pc)
 		file, _ := fn.FileLine(pc)
 		got := trimGOPATH(fn.Name(), file)
-		want := tt.want
-		if want != got {
-			t.Errorf("%v: want %q, got %q", tt.Frame, want, got)
-		}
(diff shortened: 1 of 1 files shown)


davecheney:
Thanks.

PR #80, opened 2016-08-05
Add WithStack and WithMessage tests

fabstu:
I don't know about the discussions but I like the API additions, so I took a stab at adding some tests.

Adds `testFormatCompleteCompare` as additional testing func. The new function takes a string slice as `want`, wherein stacktraces and non-stacktrace messages are discerned by `strings.ContainsAny(want[i], "\n")`. For example usage, see `TestFormatWithStack` & `TestFormatWithMessage`.

https://github.com/pkg/errors/pull/76 I do not understand the third todo, I thought causer and stack-tracer are separate interfaces so I'm a little bit confused. Is it supposed to add  `NearestCauseWithStackTrace()` or some similar function?

### DIFF
--- a/errors_test.go
+++ b/errors_test.go
@@ -218,8 +218,8 @@ func TestErrorEquality(t *testing.T) {
 		WithStack(nil),
 	}
 
-	for i := 0; i < len(vals); i++ {
-		for j := 0; j < len(vals); j++ {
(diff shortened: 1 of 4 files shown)


davecheney:
This is great, thank you!

PR #83, opened 2016-08-20
Fix the %q format for errors so it puts "" around the output

ncw:
It looks like pretty much this code was added as part of cbd18b5440315aa74a17dee2e50b67675b872b49 which appears to have been lost while merging #81


davecheney:
Thanks Nick!

Issue #96, opened 2016-12-16
Proposal: output stack trace if debugging is enabled


cyphar:
In other words, it should be possible for a user at runtime to provide a `--debug` flag to a binary and then the binary should be able to easily make any error output a stack trace.

The nicest way of doing this (IMO) would be to have something like `logrus.LogLevel` (except it would be something like `errors.EmitStack(bool)` that would determine whether `"%s"` and similar format strings will output a stack trace.

Currently an application writer has to make a decision whether they will make their program debuggable (use `%+v` everywhere) or helpful to users (use `%s` everywhere).

cyphar:
#96 is my implementation of this.

davecheney:
Thank you for your suggestion. As presented I'm not keen on adding this
because it overides the choice of code that prints the error.

Can you tell me more about your requirements, why do you need to add this
option, how are errors not being presented in a way that isnt useful for
you.

There's also th issue of this being a global flag, which is hard for
everyone to test.

cyphar:
> Thank you for your suggestion. As presented I'm not keen on adding this
> because it overides the choice of code that prints the error.
>
> Can you tell me more about your requirements, why do you need to add this
> option, how are errors not being presented in a way that isnt useful for
> you.

Let's say that I'm messing around with a program I've written, and I 
want to figure out why I'm getting a particular error. I'm using 
pkg/errors, so I know that the stack trace is stored in the error.

However, I don't want to have to modify the code and then recompile it 
in order to figure out what the stack trace is (this is software which 
has already been packaged and I've just downloaded the binary). The 
program has a --debug flag but it only gives me log output that is not 
very useful.

Unfortunately for the author (me), there is no easy way to make my 
--debug flag also output the stack trace. I might have multiple CLI 
endpoints or different logging endpoints that print the errors -- and I 
want the stacktrace to be written to *all* of them.

logrus has logrus.LogLevel(...) which lets you set the global logging 
level of the program. Why not include the same feature to pkg/errors?

> There's also th issue of this being a global flag, which is hard for
> everyone to test.

But good libraries shouldn't be printing errors to anywhere -- they 
should just be wrapping and returning them. Plus, if they just want a 
string output they can still use %s if you look at my patch[1].

The main point of this is so that I can actually make it possible for 
users of my program to provide me a stack trace of the issue, rather 
than saying "hey, can you please modify line XYZ of SOURCE.go and then 
recompile it, and then tell me what error you get now".

[...]

PR #97, opened 2016-12-16
errors: add Debug() function

cyphar:
This function allows users to use "%v" formatting everywhere, and then
dynamically decide whether they want to enable stack traces. If they
call errors.Debug(true) then "%v" formatting will return a stack trace,
otherwise it will be equivalent to "%s".

Fixes: #96
Signed-off-by: Aleksa Sarai <asarai@suse.de>

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -58,7 +58,8 @@
 //
 //     %s    print the error. If the error has a Cause it will be
 //           printed recursively
-//     %v    see %s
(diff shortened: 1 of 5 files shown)


cyphar:
I've also added tests which is basically a full copy of `format_test.go` with `Debug(true)` set in a few places.

davecheney:
Closed because #96 is closed

Issue #99, opened 2017-01-15
frame.Format: %+s prints more than "path of source file relative to the compile time GOPATH"


judwhite:
Change: https://github.com/pkg/errors/commit/7896481a535a4ecd2e5507ed4c78902d9a58ef75?diff=unified#diff-7dc02266d93e96cde6ba279eadbcac8fR62
Doc: https://github.com/pkg/errors/blob/785921b1c1ecc027064eaf6616a7a01b4c292597/stack.go#L49

Changing back to `io.WriteString(s, trimGOPATH(fn.Name(), file))` gives the expected output.

This commit was merged 6 months ago; might break people to change it now. Would you accept a PR to put the previous behavior under `%#s` and an update to the documentation?

judwhite:
@davecheney @ChrisHines Playing with this a little more I noticed `go vet` complains about `%n` (and `%#s`).

What do you think about exposing these values as methods on `Frame` with stronger guarantees than `Format`?

davecheney:
> This commit was merged 6 months ago; might break people to change it now. Would you accept a PR to put the previous behavior under %#s and an update to the documentation?

It's probably too late to change the behaviour at this point. It'd be safer to document what exists now unless it's egregiously inconsistent.

judwhite:
Current (248dadf):

```
func TestWrappedError(t *testing.T) {
	err := a()
	if st, ok := err.(stackTracer); ok {
		for _, frame := range st.StackTrace() {
			fmt.Printf("%+v\n", frame)
		}
	}
}

func a() error { return b() }
func b() error { return c() }
func c() error { return errors.Wrap(io.EOF, "unexpected eof") }
```

```
github.com/judwhite/httplog.c
        C:/Projects/go/src/github.com/judwhite/httplog/fallbackLogger_test.go:28
github.com/judwhite/httplog.b
        C:/Projects/go/src/github.com/judwhite/httplog/fallbackLogger_test.go:27
github.com/judwhite/httplog.a
        C:/Projects/go/src/github.com/judwhite/httplog/fallbackLogger_test.go:26
github.com/judwhite/httplog.TestWrappedError
        C:/Projects/go/src/github.com/judwhite/httplog/fallbackLogger_test.go:18
testing.tRunner
        C:/Go/src/testing/testing.go:610
runtime.goexit
        C:/Go/src/runtime/asm_amd64.s:2086
```

Previous (d4b5735):

```
github.com/judwhite/httplog/fallbackLogger_test.go:30
[...]

Issue #38, opened 2016-06-07
go-stack/stack and pkg/errors should support the same formatting verbs


davecheney:
At the moment, errors.Frame supports a subset of the verbs that stack.Call supports.

PR #100, opened 2017-01-15
Frame: Add File, Func, Line methods

judwhite:
Implementation of proposal in #99.

Benchmarks, for good measure:

Old:

```
$ go test -bench .
BenchmarkErrors/pkg/errors-stack-10-8            1000000              1820 ns/op             320 B/op          3 allocs/op
BenchmarkErrors/errors-stack-10-8               20000000                72.2 ns/op            16 B/op          1 allocs/op
BenchmarkErrors/pkg/errors-stack-100-8            500000              3249 ns/op             320 B/op          3 allocs/op
BenchmarkErrors/errors-stack-100-8               3000000               513 ns/op              16 B/op          1 allocs/op
BenchmarkErrors/pkg/errors-stack-1000-8           200000             11130 ns/op             320 B/op          3 allocs/op
BenchmarkErrors/errors-stack-1000-8               300000              4877 ns/op              16 B/op          1 allocs/op
PASS
ok      github.com/pkg/errors   11.294s
```

New:

```
$ go test -bench .
BenchmarkErrors/pkg/errors-stack-10-8            1000000              1825 ns/op             320 B/op          3 allocs/op
BenchmarkErrors/errors-stack-10-8               20000000                73.0 ns/op            16 B/op          1 allocs/op
BenchmarkErrors/pkg/errors-stack-100-8            500000              3229 ns/op             320 B/op          3 allocs/op
BenchmarkErrors/errors-stack-100-8               3000000               509 ns/op              16 B/op          1 allocs/op
BenchmarkErrors/pkg/errors-stack-1000-8           200000             11092 ns/op             320 B/op          3 allocs/op
BenchmarkErrors/errors-stack-1000-8               300000              4867 ns/op              16 B/op          1 allocs/op
PASS
ok      github.com/pkg/errors   11.288s
```

/cc @davecheney @ChrisHines

### DIFF
--- a/stack.go
+++ b/stack.go
@@ -46,7 +46,7 @@ func (f Frame) line() int {
 //
 // Format accepts flags that alter the printing of some verbs, as follows:
 //
-//    %+s   path of source file relative to the compile time GOPATH
+//    %+s   package name, function name, and absolute path of source file at compile time
 //    %+v   equivalent to %+s:%d
 func (f Frame) Format(s fmt.State, verb rune) {
 	switch verb {
@@ -74,6 +74,28 @@ func (f Frame) Format(s fmt.State, verb rune) {
 		io.WriteString(s, ":")
 		f.Format(s, 'd')
 	}
+}
+
+// File returns the path of the source file relative to the compile time GOPATH.
+func (f Frame) File() string {
+	pc := f.pc()
+	fn := runtime.FuncForPC(pc)
+	if fn == nil {
+		return "unknown"
+	}
+	file, _ := fn.FileLine(pc)
+	return trimGOPATH(fn.Name(), file)
+}
+
+// Func returns the function name.
+func (f Frame) Func() string {
+	name := runtime.FuncForPC(f.pc()).Name()
+	return funcname(name)
+}
+
+// Line returns the source line.
+func (f Frame) Line() int {
+	return f.line()
 }
 
 // StackTrace is stack of Frames from innermost (newest) to outermost (oldest).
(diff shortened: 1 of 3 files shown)


judwhite:
@davecheney Would you mind providing your feedback?

davecheney:
Sorry I haven't had a chance to look at this yet; I hope to have some free
time on the weekend.

judwhite:
@davecheney Friendly ping. If you want I can work on #38, although I'd still like to get specific values out of pkg/errors so users don't have to re-parse the text if they want these values on their own.

judwhite:
@davecheney @ChrisHines Could you provide guidance for a path forward or close?

davecheney:
Sorry for the extremely long delay.

I'm not sure I want to add this because it is going beyond the remit of this package. If you have the `Frame` value then you have the pc within a function and can use `runtime.FuncForPc` and friends.

If I had a choice I would not have exported `Frame`, but it was necessary to make the value implement `fmt.Formatter` _and_ still be convertible to a `uintptr`.

I'm wary about adding these methods on top of Frame because they are redundant, see above, but more seriously I'm concerned they'll be a precedent for adding more methods. This was the reason for going down the fmt.Formatter route, so the Error and Frame types would not grow a forrest of helper methods to introspect them.

judwhite:
@davecheney Thanks for the reply.

stack.go
```diff
@@ -46,7 +46,7 @@ func (f Frame) line() int {
 //
 // Format accepts flags that alter the printing of some verbs, as follows:
 //
-//    %+s   path of source file relative to the compile time GOPATH
+//    %+s   package name, function name, and absolute path of source file at compile time
```
judwhite: @davecheney btw this is the accurate documentation for `%+s` if you want to pick this in

PR #105, opened 2017-02-17
Export Causer

Xe:
There's no reason to ask people to copy-paste the `causer` definition every time they want to get the underlying cause of an error. This patch exports `Causer` as an interface type to make testing easier.


davecheney:
i would prefer not to do this as it means more people will import this package to do this assertion.

I don't want to encourage that kind of source level coupling.

PR #108, opened 2017-03-08
Use runtime.CallersFrames for StackTrace

alfred-landrum:
Given the deprecation warning about runtime.FuncForPC in:
https://docs.google.com/presentation/d/1Wcblp3jpfeKwA0Y4FOmj63PW52M_qmNqlQkNaLj0P5o/edit?usp=sharing

I'd thought I'd see what a shift to use runtime.CallersFrames would look like. I'm not yet sure of the tradeoffs for one or the other, but thought I'd share code that passes existing tests locally in case its of interest.

### DIFF
--- a/stack.go
+++ b/stack.go
@@ -8,33 +8,9 @@ import (
 	"strings"
 )
 
-// Frame represents a program counter inside a stack frame.
-type Frame uintptr
-
-// pc returns the program counter for this frame;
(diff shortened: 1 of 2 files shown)


davecheney:
Please open an issue to track this before slinging code, thank you.

alfred-landrum:
will do!

PR #113, opened 2017-04-17
Helper function for retrieving root stack trace from error

creasty:
## Why

1. Want to retrieve function name, file and line separately.
    For now, the simplest way to do so is by using Formatter as follows.
    ```go
    for _, f := range err.StackTrace() {
      file := fmt.Sprintf("%s", f)
      line, _ := strconv.ParseInt(fmt.Sprintf("%d", f), 10, 32)
      funcName := fmt.Sprintf("%n", f)
      fmt.Printf("%s at %s:%d\n", funcName, file, line)
    }
    ```
2. These is no mechanism for extracting the *root* stack trace -- the longest stack trace.

### Use case

Most feasible use case is an integration with error reporting systems such as [honeybadger](https://github.com/honeybadger-io/honeybadger-go).

https://github.com/honeybadger-io/honeybadger-go/blob/f37170201ccce33edcb44f976565698c5aeb7fc4/error.go#L13-L17

```go
type Frame struct {
	Number string `json:"number"`
	File   string `json:"file"`
	Method string `json:"method"`
}
```

## What

Create a helper function `errors.Trace` which behaves like `errors.Cause` and returns the underlying stack trace (`[]Frame`) of the error.

```go
package main

import (
	built_in_errors "errors"
	"fmt"

	"github.com/pkg/errors"
)

func foo() error {
	return built_in_errors.New("foo")
}

func bar() error {
	return errors.Wrap(foo(), "foo in bar")
}

func baz() error {
	return errors.Wrap(bar(), "bar in baz")
}

func main() {
	err := errors.Wrap(baz(), "baz in main")

	if frames := errors.Trace(err); frames != nil {
		for _, f := range frames {
			fmt.Printf("%s at %s:%d\n", f.Name(), f.File(), f.Line())
		}
	}
}
```

```hcl
$ go run main.go
bar at /Users/ykiwng/Desktop/main.go:15
baz at /Users/ykiwng/Desktop/main.go:19
main at /Users/ykiwng/Desktop/main.go:23
main at /usr/local/Cellar/go/1.7.3/libexec/src/runtime/proc.go:183
goexit at /usr/local/Cellar/go/1.7.3/libexec/src/runtime/asm_amd64.s:2086
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -267,3 +267,42 @@ func Cause(err error) error {
 	}
 	return err
 }
+
+// Trace returns the underlying stack trace of the error, if possible.
+// An error value has a stack trace if it implements the following
+// interface:
+//
+//     type stackTracer interface {
+//            StackTrace() errors.StackTrace
+//     }
+//
+// If the error does not implement StackTrace, nil will be returned.
+// If the error is nil, nil will be returned without further
+// investigation.
+func Trace(err error) (frames []Frame) {
+	if err == nil {
+		return nil
+	}
+
+	type stackTracer interface {
+		StackTrace() StackTrace
+	}
+
+	type causer interface {
+		Cause() error
+	}
+
+	for err != nil {
+		if stackTrace, ok := err.(stackTracer); ok {
+			frames = ([]Frame)(stackTrace.StackTrace())
+		}
+
+		if cause, ok := err.(causer); ok {
+			err = cause.Cause()
+		} else {
+			break
+		}
+	}
+
+	return
+}
--- a/stack.go
+++ b/stack.go
@@ -11,6 +11,11 @@ import (
 // Frame represents a program counter inside a stack frame.
 type Frame uintptr
 
+// Name returns the name of function for this Frame's pc.
+func (f Frame) Name() string {
+	return funcname(runtime.FuncForPC(f.pc()).Name())
+}
+
 // pc returns the program counter for this frame;
 // multiple frames may have the same PC value.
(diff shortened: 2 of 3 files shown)


martinheidegger:
This would also help to implement stack traces in `sentry`: https://github.com/getsentry/raven-go/issues/88

davecheney:
Please open an issue to discuss your design before sending and implantation. Thank you.

creasty:
Sorry, didn't read

> Contributing
> ...
> Before proposing a change, please discuss your change by raising an issue.

davecheney:
Thank you. It is important to discuss the problem you have before proposing
a solution.

gregwebs:
I [added this functionality as `GetStackTracer`](https://github.com/pingcap/errors/blob/master/errors.go#L180) on [our fork](https://github.com/pingcap/errors/blob/master/errors.go#L180).

creasty:
Actually, I've created a better solution at https://github.com/srvc/fail and stopped using pkg/errors.

c.f., Create a virtual stack trace by merging multiple stack traces
https://github.com/srvc/fail/pull/13

gregwebs:
@creasty I saw svc/fail. I really like what the package offers and particularly the idea of being able to merge stack traces. However, I am concerned about the potential for worse performance with all the stack trace creation and I am also working towards a different place in the design space by using our fork of pkg/errors as a base and building a [package for error codes](https://github.com/pingcap/pd/blob/ee73eb5270102084ac2b3c487f6b84b06c7f2a98/pkg/error_code/error_code.go).

PR #114, opened 2017-05-04
Fixed #102, a message formatting bug.

### LINKED ISSUE #102: WithMessage appends message after associated stack traces instead of before
After amending an error using errors.WithMessage, printing the error with `%+v` prints the extra information after the stack trace instead of with the original message. For example:

```go
package main

import (
	"log"

	"github.com/pkg/errors"
)

func a() error {
	return errors.New("error details")
}

func b() error {
	if err := a(); err != nil {
		return errors.WithMessage(err, "additional details")
	}
	return nil
}

func main() {
	if err := b(); err != nil {
		log.Fatalf("%+v", err)
	}
}
```

This prints:

```
2017/02/06 14:34:13 error details
main.a
	.../example.go:10
main.b
	.../example.go:14
main.main
	.../example.go:21
runtime.main
	/usr/local/Cellar/go/1.7.4_2/libexec/src/runtime/proc.go:183
runtime.goexit
	/usr/local/Cellar/go/1.7.4_2/libexec/src/runtime/asm_amd64.s:2086
additional details
exit status 1
```

As mentioned in #75, it seems like it would be more helpful to include the additional information with original error message. When using `%v` with the same example, it prints messages in this way now:

```
2017/02/06 14:33:32 additional details: error details
exit status 1
```

mrhwick:
Fixes https://github.com/pkg/errors/issues/102

With these changes, when using a plain `withMessage()`, we'll print the additional message in front of the original error message, above the stack trace. This maintains the correct ordering when an error is being `Wrap()`'d, printing the message + second stack trace after the first message + stack trace.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -182,8 +182,9 @@ func Wrap(err error, message string) error {
 		return nil
 	}
 	err = &withMessage{
-		cause: err,
-		msg:   message,
+		cause:              err,
+		msg:                message,
+		stackedOnWithStack: true,
 	}
 	return &withStack{
 		err,
@@ -199,8 +200,9 @@ func Wrapf(err error, format string, args ...interface{}) error {
 		return nil
 	}
 	err = &withMessage{
-		cause: err,
-		msg:   fmt.Sprintf(format, args...),
+		cause:              err,
+		msg:                fmt.Sprintf(format, args...),
+		stackedOnWithStack: true,
 	}
 	return &withStack{
 		err,
@@ -215,14 +217,16 @@ func WithMessage(err error, message string) error {
 		return nil
 	}
 	return &withMessage{
-		cause: err,
-		msg:   message,
+		cause:              err,
+		msg:                message,
+		stackedOnWithStack: false,
 	}
 }
 
 type withMessage struct {
-	cause error
-	msg   string
+	cause              error
+	msg                string
+	stackedOnWithStack bool
 }
 
 func (w *withMessage) Error() string { return w.msg + ": " + w.cause.Error() }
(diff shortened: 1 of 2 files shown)


davecheney:
Not LGTM sorry. Do we need a field on withMessage? Can't you do  type assertion instead?

mrhwick:
Essentially, there are two use cases we want to meet here. 

1. In the case where a `withMessage` is added to the chain independently of a `withStack`, we want to print them separately. In this case, we print the message prepended to the entire error format, and the additional stack appended to the error format.
2. In the case where a `Wrap()` call adds a `withMessage` first, and then immediately adds a `withStack`, we want to keep them together because they are related. In this case, we would want to print the entire error format, then the `withMessage` and then the `withStack`. This will keep the message associated with its related stack.

If you use type assertion to accomplish this, you might do the type assertion in the `withStack` `Format()` function.

```go
func (w *withStack) Format(s fmt.State, verb rune) {
	switch verb {
	case 'v':
		if s.Flag('+') {
			withMessageWrapper, ok := w.Cause().(*withMessage) // Type Assertion
			if ok { // Run this whenever we have a withMessage next in the chain
				fmt.Fprintf(s, "%+v\n", withMessageWrapper.Cause())
				io.WriteString(s, withMessageWrapper.msg)
				w.stack.Format(s, verb)
			} else { // Run this for anything else next in the chain
				fmt.Fprintf(s, "%+v", w.Cause())
				w.stack.Format(s, verb)
			}
			return
		}
		fallthrough
	case 's':
		io.WriteString(s, w.Error())
	case 'q':
		fmt.Fprintf(s, "%q", w.Error())
	}
}```

 This introduces an issue that from that scope there is no way to determine whether the `withStack` wraps a `withMessage` as a part of a `Wrap()` call. It could be a `withMessage` that should be kept separate, or it could be a `withMessage` that should be kept with its related `withStack`.

The scenario can easily occur where a pre-existing `withMessage` is placed in the chain before being passed to a `WithStack()` call. In these cases, the preexisting `withMessage` is indistinguishable from a `withMessage` added by calling `Wrap()`, and will be assumed to be related to the earlier `withStack` in the chain.

Maybe that is a viable solution? Is there a scenario where a `WithMessage()` followed next by `WithStack()` wouldn't be assumed to be related to one another?

hanzei:
What is the status of this PR? Does the comment above helped the discussion?

PR #115, opened 2017-05-04
Add doc comment for exported Format func

mrhwick:
https://goreportcard.com/report/github.com/pkg/errors

Currently, there is only one issue on the report for github.com/pkg/errors above. This PR fixes that issue, which was an undocumented, but exported, Format function.


davecheney:
Thanks

PR #116, opened 2017-05-18
message getter added

tugbadartici:
Message getter added


davecheney:
Please follow the contribution procedure outlined in the readme and open an issue before sending a PR.

PR #118, opened 2017-06-05
Add WithMessagef function

chemicL:
WithMessagef utility function to accompany WithMessage, similar to
exiting Wrapf function accompanying Wrap.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -220,6 +220,18 @@ func WithMessage(err error, message string) error {
 	}
 }
 
+// WithMessagef annotates err with the format specifier.
+// If err is nil, WithMessagef returns nil.
+func WithMessagef(err error, format string, args ...interface{}) error {
+	if err == nil {
+		return nil
+	}
+	return &withMessage{
+		cause: err,
+		msg: fmt.Sprintf(format, args...),
+	}
+}
+
 type withMessage struct {
 	cause error
 	msg   string
--- a/errors_test.go
+++ b/errors_test.go
@@ -198,6 +198,32 @@ func TestWithMessage(t *testing.T) {
 	}
 }
 
+func TestWithMessagefNil(t *testing.T) {
+	got := WithMessagef(nil, "no error")
+	if got != nil {
+		t.Errorf("WithMessage(nil, \"no error\"): got %#v, expected nil", got)
+	}
+}
+
+func TestWithMessagef(t *testing.T) {
+	tests := []struct {
+		err     error
+		message string
+		want    string
+	}{
+		{io.EOF, "read error", "read error: EOF"},
+		{WithMessagef(io.EOF, "read error without format specifier"), "client error", "client error: read error without format specifier: EOF"},
+		{WithMessagef(io.EOF, "read error with %d format specifier", 1), "client error", "client error: read error with 1 format specifier: EOF"},
+	}
+
+	for _, tt := range tests {
+		got := WithMessagef(tt.err, tt.message).Error()
+		if got != tt.want {
+			t.Errorf("WithMessage(%v, %q): got: %q, want %q", tt.err, tt.message, got, tt.want)
+		}
+	}
+}
+
 // errors.New, etc values are not expected to be compared by value
 // but the change in errors#27 made them incomparable. Assert that
 // various kinds of errors have a functional equality operator, even


chemicL:
Cheers, @JReyLBC. Hopefully @davecheney accepts this :-)

davecheney:
Hello,

To give  you some visibility here. You didn't follow the contribution process, you just threw some code over the wall. That's not cool. 

The reason I ask people to submit an issue before writing code is so we discuss it first, then if necessary a change is made.

I understand why you are suggesting adding this feature, but my strong desire for this package is to keep it small. What I am doing is surveying the importers of this package, http://godoc.org/github.com/pkg/errors?importers. As you can see, there are a lot, thousands, and while I won't survey every single use, I do want to see how many are combining `fmt.Sprintf` with `errors.WithMessage`. If the number is small, then I will no accept this change. If it is clear that there is a pattern of people having to reach for `fmt.Sprintf` to use `WithMessage`, then I'll merge this change.

In future, please follow the contribution guidelines and discuss your change _before_ starting to code. 

Thank you.

dragonsinth:
We had to write a helper to do this when we went through and audited our existing uses of errors.Wrapf().  We had too much stack trace explosion, and it would have been nice to trivially convert this to errors.WithMessagef().  We wrote our own helper because it was too much manual effort to jam a fmt.Sprint() into the middle of each call-- it's not amenable to simple search/replace.

vbezhenar:
I've found that for my code I would only use WithMessage with parameters. Here's my reasoning:

1. I don't want to add another meaningless stacktrace, it means that `err` comes from my code with already meaningfull stacktrace.
2. I don't want to just add string message: stacktrace already contains function names and static string won't add much.
3. What I actually do want to add is some variable values to improve context. Stacktrace doesn't include variable values, so they might add important information about error.

In my code I always using `WithMessage` and `fmt.Sprintf` together. Not a big deal, but collapsing them into one function is only logical and currently it's inconsistent because every other function has analog with formatting.

dragonsinth:
Well put @vbezhenar!  That captures my feelings exactly.

natefinch:
I get lazy and use Wrapf when what I really want is WithMessagef. I'm sure I'm not the only one.

aviau:
>I don't want to add another meaningless stacktrace, it means that err comes from my code with already meaningfull stacktrace.

I have the same feeling.

However, how can I know if there is a stack trace already?

I have created new WithMessage/WithMessagef methods in my code:
```golang
import (
	"fmt"
	pkgerrors "github.com/pkg/errors"
)

//WithMessage annotates an error with a message. Unlike github.com/pkg/errors,
//this implementation also annotates the error with a stack trace if there
//is not already one.
func WithMessage(err error, message string) error {

	type stackTracer interface {
		StackTrace() pkgerrors.StackTrace
	}

	// Try to find a StackTracer down the error cause chain.
	current := err
	for current != nil {
		if _, ok := current.(stackTracer); ok {
			break
		}
		current = nextCause(current)
	}

	// The error is not yet annotated with a stack trace, Wrap it.
	if current == nil {
		err = pkgerrors.Wrap(err, message)
	}

	return pkgerrors.WithMessage(err, message)
}

//WithMessagef annonates an error with a formatted message
func WithMessagef(err error, format string, args ...interface{}) error {
	if err == nil {
		return nil
	}
	message := fmt.Sprintf(format, args...)
	return WithMessage(err, message)
}

//nextCause returns the direct cause of the error
func nextCause(err error) error {
	type causer interface {
		Cause() error
	}

	if causer, ok := err.(causer); ok {
		return causer.Cause()
	}

	return nil
}
```

ondrej-fabry:
> I do want to see how many are combining `fmt.Sprintf` with `errors.WithMessage`. If the number is small, then I will no accept this change. If it is clear that there is a pattern of people having to reach for `fmt.Sprintf` to use `WithMessage`, then I'll merge this change.

@davecheney It won't be that simple to know if people would use it by searching this. As other already mentioned, people often tend to look for simpler solutions than repeating the same thing all over their code.

```go
if err != nil {
    return errors.WithMessage(err, fmt.Sprintf("failed to add %v", x))
}
```
vs.
```go
if err != nil {
    return errors.WithMessagef(err, "failed to add %v", x)
}
```

And I believe you wouldn't use the first variant either if it occurred quite handful of times in your code, but would implement some helper function for it. Or even *wrap* the entire **errors** package with custom implementation like @aviau did.

Please consider accepting this.

davecheney:
Thank you for this PR.

dragonsinth:
Yay!

Issue #75, opened 2016-07-27
Why call callers() on every wrap?


cep21:
Every time an error is wrapped, callers() is called.  This seems wasteful to me.  I think the root error's stack trace is the most important and almost always what I want.

I think if the error that is being wrapped implements `stackTrace interface`, that error's stack trace should be used instead, and the call to callers() can be skipped. For sure, any stack traces in errors between the top and bottom of the error stack are ignored.

I think my optimal format message would include the wrapped msg's all down the various error wraps with a stack trace at the bottom.

ChrisHines:
I agree with @cep21 in general. The only scenario where multiple stack traces *might* be useful is when an error is passed between goroutines (and thus jumps stacks) before getting handled.

davecheney:
You only need to call Wrap when you have extra context to return.

I'm not concerned about the cost of calling Wrap in the error path.

Please see this blog post.
http://dave.cheney.net/2016/06/12/stack-traces-and-the-errors-package

cep21:
The way the code is now, all the stack traces of wraps get put into the output.  Isn't that excessive?

> If you interact with a package from another repository, consider using errors.Wrap or errors.Wrapf to establish a stack trace at that point

If I encourage myself and coworkers to use the errors package, their returns would already include stack traces.  Maybe.  It's a bit too much for me to know which libraries I'm using have stacks in their errors and which don't.  That goes away from "errors are opaque" territory.

I can't see much practical use from stack traces of errors wrapped in the middle of the chain.

benma:
I agree that the behaviour is not the expected one. If I do `errors.Wrap(err, "more context")`, and `err` already has a stacktrace, I don't want to add more stacktrace to it, but just prefix `more context: ` to the output.

If you use `errors.Wrap()` a number of times in a row (once per stack frame), the vast majority of the output is duplication, and the only interesting stack trace is the longest one.

Please consider changing it.

ChrisHines:
We should have control over this once https://github.com/pkg/errors/pull/76 gets merged.

davecheney:
> I agree that the behaviour is not the expected one. If I do errors.Wrap(err, "more context"), and err already has a stacktrace, I don't want to add more stacktrace to it, but just prefix more context: to the output.

Why not ? I can equally see a new too add a strack trace for each wrap. For example
```go
err := <- ch
if err != nil {
       return errors.Wrap(err, "async request failed")
}
...
```
[...]

PR #122, opened 2017-06-17
#75 Wrap() add withStack only if no cause with stack present

nattawitc:


natefinch:
Yes please! I don't want to have to figure out if an incoming error has a stacktrace yet or not.  I just want to ensure it has one.... but I definitely don't want to add another stack trace on top of the original one, since this is often a mostly duplicated stack trace (i.e. one level up).

abraithwaite:
Copying my comment here for people to see.  This isn't an ideal solution but it works well enough:

> Came to this thread with the same issue everyone else is having.
> This is how I solved it without forking this package or requiring additional dependencies.
> https://github.com/abraithwaite/go-examples/blob/master/errdive.go#L9-L30
> You don't get the extra messages, but if you have the stack do you really need those?

cstockton:
Could this use a technique similar to causer.Cause with assertion for stackTracer? i.e.:

```
func isStackTracer(err error) (ok bool) {
	if _, ok = err.(stackTracer); ok {
		return
	}
	if err, ok = err.(causer); ok {
		return isStackTracer(err)
	}
	return
}
```

gregwebs:
This change is problematic because it changes backwards compatibility and somebody else actually wants it this way. On our [fork of pkg/errors](https://github.com/pingcap/errors) we [added new functions](https://github.com/pingcap/errors/pull/2/files#diff-dd25e0e0a594c3f3592c1c717039b85eR164) `Annotate/Annotatef` and consider `Wrap/Wrapf` deprecated.

gregwebs:
@develar this is one of several issues addressed in our fork. In this case, the forks are interoperable: one package can use pkg/errors and another package can use pingcap/errors. If you don't want the little bit of code bloat you can tell your package tool to use pingcap/errors as pkg/errors in your project. This does mean that pingcap/errors has to continue tracking any changes in pkg/errors, but that has been easy so far since there haven't been any.

hanzei:
Can this be merged?

PR #134, opened 2017-09-19
Message(error) function, to make it easy to assert top most error

adzr:
Hi, I thought about adding a function to expose the message that is passed when wrapping an error since it may help checking which error has returned during the handling, well..to make more clear, here is a example of what I got stuck into:
```golang
package processor

import (
  "github.com/pkg/errors"
)

func Process() error {
  if err := validateState(); err != nil {
    return errors.Wrap(err, "validation failed")
  }

  if err := processData(); err != nil {
    return errors.Wrap(err, "processing failed")
  }

  return nil
}
```
Now consider that I'm calling function Process() from outside the package 'processor':
```golang
func main() {
  .
  .
  .
  if err != processor.Process(); err != nil {
    // Here I need to check if err variable contains a validation error or a processing error
    // so I can handle each accordingly.
  }
}
```
The problem here is that I can't determine which error is which without a minor - still not ideal - string processing, since the error message that I used in wrapping is suffixed with an arbitrary cause message with no direct way to split each.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -225,8 +225,9 @@ type withMessage struct {
 	msg   string
 }
 
-func (w *withMessage) Error() string { return w.msg + ": " + w.cause.Error() }
-func (w *withMessage) Cause() error  { return w.cause }
+func (w *withMessage) Error() string   { return w.msg + ": " + w.cause.Error() }
+func (w *withMessage) Cause() error    { return w.cause }
+func (w *withMessage) Message() string { return w.msg }
 
 func (w *withMessage) Format(s fmt.State, verb rune) {
 	switch verb {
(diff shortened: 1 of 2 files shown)


davecheney:
Thank you for sending this PR. The contribution policy of this repo states that we discuss new features before sending code. Please raise an issue to discuss your change and reference it here. 

Tl;dr thanks but no thanks, if you need to inspect the error then your should write your own ValidationError type then wrap it with errors.WithStack.

adzr:
My pleasure. your second statement indicates that I shouldn't bother raising an issue regarding this, am I right?

davecheney:
I'd welcome the discussion in an issue but as it stands I don't believe changing the errors package is the solution to your problem.

adzr:
Thanks for the feedback, I might consider an alternative solution then.

puellanivis:
As an idea to assist you in an alternative solution, if you need to distinguish validation errors from processing errors, then you should likely split the two functions apart away from a single verify-and-process step.

i.e.: call a `Validate() error` first, handle validation errors as necessary, then call a separate `Process() error` function and handle the processing errors as necessary.

More code? Yes, but it reads much more clearly, and splits semantic elements. (That is, it avoids an implicit Validate() call that isn’t immediately apparent from the Process() function name. Which I like to term “magic”.)

adzr:
Thanks for your input, actually the code snippet provided is something that I came up with on the spot while writing, however, the real scenario follows a clean architecture pattern where the business domain model functions may return more than one possible error type, it could be business validation, adapter IO or whatever.
In the service layer has to handle each error type accordingly and publish the reason of failure back to the caller in a more abstract form.
So I decided to stick with my fork temporarily until I find a better solution, and if I do I'll try to remember to post it back here :)
I really appreciate the help, thanks again.

PR #137, opened 2017-09-29
Fix doc comment for exported Format func

### LINKED ISSUE #136: Inconsistency between documentation and code
I found a little inconsistency between the documentation and the actual implementation.

The documentation says that if a `Frame` is formatted with `%+s` the produced string is the "path of source file relative to the compile time GOPATH". [link](https://github.com/pkg/errors/blob/master/stack.go#L49)

However, according to the [implementation](https://github.com/pkg/errors/blob/master/stack.go#L62) it also contains the function name.

Is this difference intended or not?

ihrwein:
Fixes #136

Signed-off-by: Tibor Benke <ihrwein@gmail.com>

### DIFF
--- a/stack.go
+++ b/stack.go
@@ -46,7 +46,8 @@ func (f Frame) line() int {
 //
 // Format accepts flags that alter the printing of some verbs, as follows:
 //
-//    %+s   path of source file relative to the compile time GOPATH
+//    %+s   function name and path of source file relative to the compile time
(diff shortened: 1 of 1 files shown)


gjnoonan:
:+1: LGTM

Can this be merged?

davecheney:
LGTM. Thank you sorry for the delay

PR #146, opened 2018-01-27
Remove deadcode

philoserf:
- The only place `TrimGOPATH` was used was its test
- I found one old closed issue reference at #99, and
- the code was introduced in pull request #37 

I found no reason to keep the otherwise dead code.


davecheney:
SGTM. Thanks.

PR #155, opened 2018-05-05
export StackTracer and ErrorWithStackTrace

### LINKED ISSUE #103: Proposal: How to use Type system to force use of decorated errors
First off: thanks for a really great library and hope this is the right place for this.

Within the confines of my particular application I'd really like to make sure that *ALL* errors have been decorated with a stack trace. I feel like the best way to do this would be with the type system, perhaps:

```go
func MyRecordCreationFunction(some args...) (record, errors.ErrorWithStack) {
   record, err := orm.CreateObj("...")
   if err != nil {
     return nil, errors.WithStack(err)
   }

   err = record.hydrate()
   if err != nil {
     return nil, err // <- fails to compile, type error does not implement method StackTrace() ...
   }

   return record, nil
}
```

Unfortunately this is quite challenging with the types that this package decides to export right now. Is there another way to achieve this? Or is this worthy of a PR?

Thanks for your time.

aviau:
Closes #103

This allows to use the type system to ensure that a function returns an error with a stack trace.

For example:
```golang
func getNumber() (int, ErrorWithStackTrace)
```

AFAIK there is no other way to do this than exporting this interface.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -96,9 +96,18 @@ import (
 	"io"
 )
 
+type StackTracer interface {
+	StackTrace() StackTrace
+}
+
+type ErrorWithStackTrace interface {
+	error
+	StackTracer
+}
+
 // New returns an error with the supplied message.
 // New also records the stack trace at the point it was called.
-func New(message string) error {
+func New(message string) ErrorWithStackTrace {
 	return &fundamental{
 		msg:   message,
 		stack: callers(),
@@ -108,7 +117,7 @@ func New(message string) error {
 // Errorf formats according to a format specifier and returns the string
 // as a value that satisfies error.
 // Errorf also records the stack trace at the point it was called.
-func Errorf(format string, args ...interface{}) error {
+func Errorf(format string, args ...interface{}) ErrorWithStackTrace {
 	return &fundamental{
 		msg:   fmt.Sprintf(format, args...),
 		stack: callers(),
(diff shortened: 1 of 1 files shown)


davecheney:
Sorry. This package does not provide the guarantee that you seek. 

I encourage you to fork this package and make the modifications you require.

aviau:
@CreatCodeBuild 

Your proposed code does compile without adding type assertions in the code.

Errors returned by github.com/pkg/errors implement the StackTracer interface, but the methods in the package do not return StackTracers. We must perform type assertion.

This does not have to be done at runtime.

@davecheney Would you consider another approach?

For example, would you change the method signatures so that they return the following interface:

```
type ErrorWithStackTrace interface {
    error
    StackTrace() fmt.Formatter
}
```

Or something similar, such as a Stringer, or anything else that wouldn't necessarily expose the details of the StackTrace implementation.

davecheney:
@aviau nope sorry, the return type for the errors package must be `error`. Consider this piece of code:
```
err := errors.New("foo") // returns errors.ErrorWithStackTrace
c, err := net.Listen("...") // won't compile
```

davecheney:
Please continue the discussion on #103.

aviau:
Right, I had not considered this. I think that this makes it clear that there is no way to modify the package in a way that is not breaking and would allow me doing what I want.

I'll close this PR.

thank you for your responses, I love this package btw.

aviau:
I have implemented this in a new package:
 - https://github.com/aviau/errors

PR #159, opened 2018-06-12
Add versions of the wrapper functions that skip extra stack frames

ungerik:
In cases where the wrapper functions of this package are called in other utility functions, the stack frame of those utility functions is included. This often doesn't add any value, it just distracts from debugging the real error.

A typical use case is a deferred call to a utility function, that adds meta information about the calling function to any error returned from that function.

By adding versions of the wrap functions that take a skip int argument, utility functions can skip their own stack frames.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -101,7 +101,7 @@ import (
 func New(message string) error {
 	return &fundamental{
 		msg:   message,
-		stack: callers(),
+		stack: callers(0),
 	}
 }
 
@@ -111,7 +111,7 @@ func New(message string) error {
 func Errorf(format string, args ...interface{}) error {
 	return &fundamental{
 		msg:   fmt.Sprintf(format, args...),
-		stack: callers(),
+		stack: callers(0),
 	}
 }
 
@@ -147,7 +147,19 @@ func WithStack(err error) error {
 	}
 	return &withStack{
 		err,
-		callers(),
+		callers(0),
+	}
+}
+
+// WithStackSkip annotates err with a stack trace at the point WithStackSkip was called minus skip stack frames.
+// If err is nil, WithStackSkip returns nil.
+func WithStackSkip(skip int, err error) error {
+	if err == nil {
+		return nil
+	}
+	return &withStack{
+		err,
+		callers(skip),
 	}
 }
 
@@ -187,7 +199,24 @@ func Wrap(err error, message string) error {
 	}
 	return &withStack{
 		err,
-		callers(),
+		callers(0),
+	}
+}
+
+// WrapSkip returns an error annotating err with a stack trace
+// at the point WrapSkip is called minus skip stack frames, and the supplied message.
+// If err is nil, WrapSkip returns nil.
+func WrapSkip(skip int, err error, message string) error {
+	if err == nil {
+		return nil
+	}
+	err = &withMessage{
+		cause: err,
+		msg:   message,
+	}
+	return &withStack{
+		err,
+		callers(skip),
 	}
 }
 
@@ -204,7 +233,24 @@ func Wrapf(err error, format string, args ...interface{}) error {
 	}
 	return &withStack{
 		err,
-		callers(),
+		callers(0),
+	}
+}
+
+// WrapfSkip returns an error annotating err with a stack trace
+// at the point WrapfSkip is called minus skip stack frames, and the format specifier.
+// If err is nil, WrapfSkip returns nil.
+func WrapfSkip(skip int, err error, format string, args ...interface{}) error {
+	if err == nil {
+		return nil
+	}
+	err = &withMessage{
+		cause: err,
+		msg:   fmt.Sprintf(format, args...),
+	}
+	return &withStack{
+		err,
+		callers(skip),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


gregwebs:
I [added this capability](https://github.com/pingcap/errors/pull/4/files#diff-7dc02266d93e96cde6ba279eadbcac8fR168) as a function `NewStack(skip int) StackTracer` on [our fork](https://github.com/pingcap/errors/pull/4/files#diff-7dc02266d93e96cde6ba279eadbcac8fR168).

gregwebs:
I am using uber-go/multierr to combine errors

puellanivis:
The scope of this proposed changes has changed, please update at least the PR title and summary.

combination.go
```diff
@@ -0,0 +1,125 @@
...
+	"io"
+	"strings"
+)
+
+// Combination combines multiple errors into one.
+// The Error method returns the strings from the individual Error methods
+// joined by the new line character '\n'.
+// Combination always has at least one error and returns the first error
+// as result of the Cause method.
+type Combination struct {
```
puellanivis: You will notice that the current master API does not share any datatypes except those dealing with stack frames. This is an intentional choice, as typed errors often lead to a lot of issues.

combination.go
```diff
@@ -0,0 +1,125 @@
...
+func Combine(errs ...error) error {
+	switch len(errs) {
+	case 0:
+		return nil
+	case 1:
+		return WithStackSkip(1, errs[0])
+	}
+
+	var lastNotNil error
+	numNotNil := 0
```
puellanivis: `var numNotNil int`

combination.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+	switch numNotNil {
+	case 0:
+		return nil
+	case 1:
+		return WithStackSkip(1, lastNotNil)
+	}
+
+	if numNotNil < len(errs) {
```
puellanivis: `append` does a very good job of amortizing allocation costs. As such, it is strongly recommended that one simplify their code rather than attempt to "squeeze every ounce of performance" out of this.

This whole function would be much more readable as:
```
var collect []error
for _, err := range errs {
  if err != nil {
    collect = append(collect, err)
  }
}
switch len(collect) {
case 0;
  return nil
case 1:
  return WithStackSkip(1, collect[0])
}
return &combination{
  errs: errs,
  stack: callers(0),
}
```

combination.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// Uncombine returns multible errors if err is a Combination type,
+// or the passed single err if was not a Combination type,
+// or nil if the passed error was nil.
+func Uncombine(err error) []error {
+	if err == nil {
+		return nil
+	}
+	if combo, ok := err.(*Combination); ok {
```
puellanivis: Do not test for type, test for behavior.  The unexported version of your type `combination` implements `Errors() []error` and if you use that in your type assertion, then this package would be able to `Uncombine` errors from other packages as well, for instance, from forks.

combination.go
```diff
@@ -0,0 +1,125 @@
...
+		return combo.errs
+	}
+	return []error{err}
+}
+
+func (c *Combination) Error() string {
```
puellanivis: If we export this type, then I can do something like: `new(errors.Combination).Error()` which here would produce an empty-string, which is not ideal.

If the type were not exported, then there is a much stronger chance that we can ensure that short of someone using `reflect` that we will not encounter a zero-value `Combination`.

combination.go
```diff
@@ -0,0 +1,125 @@
...
+	return b.String()
+}
+
+func (c *Combination) Errors() []error {
+	return c.errs
+}
+
+func (c *Combination) Append(err error) {
+	if err != nil {
+		c.errs = append(c.errs, WithStackSkip(1, err))
```
puellanivis: The `Combine` function does not attach a Stack to every error. Attaching the stack to the error in this function, makes things inconsistent.

combination.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func (c *Combination) Cause() error {
+	return Cause(c.errs[0])
```
puellanivis: Given `new(errors.Combination).Cause()` we get: `panic: array index out of range`

An argument can additionally be made that this is not the intended semantics of a `Cause` on a `combination`. Because maybe someone is putting the ultimate cause at the end of the `combination` not the start?

combination_test.go
```diff
@@ -0,0 +1,63 @@
+package errors
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
puellanivis: https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures Do not use asserts. Use useful test failures.

errors.go
```diff
@@ -147,7 +147,19 @@ func WithStack(err error) error {
...
 	return &withStack{
 		err,
-		callers(),
+		callers(0),
+	}
+}
+
+// WithStackSkip annotates err with a stack trace at the point WithStackSkip was called minus skip stack frames.
+// If err is nil, WithStackSkip returns nil.
+func WithStackSkip(skip int, err error) error {
```
puellanivis: What if I call this function with `skip == -4`?

combination.go
ungerik: Sorry, wasn't aware that this PR was still open, should be closed. My further additions were meant as permanent fork.

PR #161, opened 2018-06-20
Wrapf: remove duplicated code

grrtrr:
Wrap and Wrapf have the same behaviour and nearly identical function bodies.
This removes the duplicate code by expressing Wrapf in terms of Wrap.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -195,17 +195,7 @@ func Wrap(err error, message string) error {
 // at the point Wrapf is call, and the format specifier.
 // If err is nil, Wrapf returns nil.
 func Wrapf(err error, format string, args ...interface{}) error {
-	if err == nil {
-		return nil
-	}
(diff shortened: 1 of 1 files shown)


grrtrr:
This broke the test.

puellanivis:
What went wrong here (I tried thinking about this before myself) was that the `callers()` call expects to be called directly from the Wrap/Wrapf function. When you are writing Wrapf in terms of Wrap, `callers` is now being called one deeper in the stack, which it fails to skip over, and thus this.

I’ve basically left all refactoring of the common code alone as a result, as any such work would have to include the `callers()` in the proper frame and 🤷‍♀️ at that point you’re overcomplicating the code just to remove some extremely simplistic code.

grrtrr:
Thank you for the explanation. It was a good reminder to run tests first; they are well done.

PR #165, opened 2018-08-17
Allow version-tagged filepath in tests

thsnr:
Go 1.11 will introduce support for [versioned modules](https://github.com/golang/go/wiki/Modules). When downloading module dependencies, `go get` will add a version tag to the directory where the package is checked out (e.g., `github.com/pkg/errors@v0.8.0`).

It is also recommended as a best practice to run `go test all` before release, which will run tests for the package being released and all dependencies to check compatibility. Currently this fails because the tests are not happy to be checked out in a version-tagged directory.

```
$ cat go.mod 
module github.com/thsnr/gomod

require github.com/pkg/errors v0.0.0-00000000000000-816c9085562cd7ee03e7f8188a1cfd942858cded

$ go1.11rc1 test all
?   	github.com/thsnr/gomod	[no test files]
...
--- FAIL: TestFrameFormat (0.00s)
    format_test.go:379: test 2: line 2: fmt.Sprintf("%+s", err):
         got: "github.com/pkg/errors.init\n\t/home/tiit/go/pkg/mod/github.com/pkg/errors@v0.0.0-00000000000000-816c9085562cd7ee03e7f8188a1cfd942858cded/stack_test.go"
        want: "github.com/pkg/errors.init\n\t.+/github.com/pkg/errors/stack_test.go"
    format_test.go:379: test 12: line 2: fmt.Sprintf("%+v", err):
         got: "github.com/pkg/errors.init\n\t/home/tiit/go/pkg/mod/github.com/pkg/errors@v0.0.0-00000000000000-816c9085562cd7ee03e7f8188a1cfd942858cded/stack_test.go:9"
        want: "github.com/pkg/errors.init\n\t.+/github.com/pkg/errors/stack_test.go:9"
...
```

This pull request simply changes the tests to allow a version tag. I purposefully left the version regular expression vague (`@v[^/]+`) to keep it short – these tests should not concern themselves with correctness of the tag – and inlined it to avoid changing any line numbers.

### DIFF
--- a/format_test.go
+++ b/format_test.go
@@ -27,7 +27,7 @@ func TestFormatNew(t *testing.T) {
 		"%+v",
 		"error\n" +
 			"github.com/pkg/errors.TestFormatNew\n" +
-			"\t.+/github.com/pkg/errors/format_test.go:26",
+			"\t.+/github.com/pkg/errors(@v[^/]+)?/format_test.go:26",
 	}, {
 		New("error"),
 		"%q",
@@ -57,7 +57,7 @@ func TestFormatErrorf(t *testing.T) {
 		"%+v",
 		"error\n" +
 			"github.com/pkg/errors.TestFormatErrorf\n" +
-			"\t.+/github.com/pkg/errors/format_test.go:56",
+			"\t.+/github.com/pkg/errors(@v[^/]+)?/format_test.go:56",
 	}}
 
 	for i, tt := range tests {
@@ -83,7 +83,7 @@ func TestFormatWrap(t *testing.T) {
 		"%+v",
 		"error\n" +
 			"github.com/pkg/errors.TestFormatWrap\n" +
-			"\t.+/github.com/pkg/errors/format_test.go:82",
+			"\t.+/github.com/pkg/errors(@v[^/]+)?/format_test.go:82",
 	}, {
 		Wrap(io.EOF, "error"),
 		"%s",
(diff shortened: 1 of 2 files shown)


davecheney:
Thanks for your PR. I want to think about why this is breaking in the first
place before I apply this fix. I'm sure its correct, but I wonder if it's
papering over a broader problem with assuming the prefix is stable.

ChrisHines:
FWIW, I was just playing with adding module support to github.com/go-stack/stack and encountered the same problem with the tests in that package. In particular, if I work in module mode outside GOPATH with go1.11rc1 then the logic [to trim the compile time GOPATH from source file paths](https://github.com/go-stack/stack/blob/master/stack.go#L248) no longer works because _there is no compile time GOPATH any more_. That breaks the `%+s` and `%+v` format tests.

I don't have a great solution yet.

thsnr:
Ah yes, I forgot to mention that this problem only occurs when operating in "module-aware mode" as opposed to "GOPATH mode".

@davecheney I do not think I fully understand the GOPATH trimming problem. While yes, `github.com/go-stack/stack` does trim the GOPATH from filenames before printing, I see no such behavior here in `github.com/pkg/errors` (although promised in https://github.com/pkg/errors/blob/master/stack.go#L49).

I even ran a small test-case:

```go
package main

import (
        "fmt"

        "github.com/pkg/errors"
)

type stackTracer interface {
}

func main() {
        err := errors.New("stack")
        fmt.Printf("%+s\n\n", err.(stackTracer).StackTrace()[0])
        fmt.Printf("%+v\n", errors.New("stack"))
}
```

Result in "module-aware mode":
```
main.main
        /tmp/gomod/main.go

stack
main.main
        /tmp/gomod/main.go:16
runtime.main
        /home/tiit/sdk/go1.11rc1/src/runtime/proc.go:201
runtime.goexit
        /home/tiit/sdk/go1.11rc1/src/runtime/asm_amd64.s:1333
```

Result in "GOPATH mode":
```
main.main
        /tmp/gopath/src/github.com/thsnr/stack/main.go

stack
main.main
        /tmp/gopath/src/github.com/thsnr/stack/main.go:16
runtime.main
        /home/tiit/sdk/go1.11rc1/src/runtime/proc.go:201
runtime.goexit
        /home/tiit/sdk/go1.11rc1/src/runtime/asm_amd64.s:1333
```

(Output when using Go 1.10 was same as GOPATH mode, just with a different GOROOT.)

No GOPATH trimming is happening, it simply uses the output of [Func.FileLine](https://golang.org/pkg/runtime/#Func.FileLine).

If I could understand the situation better, I would be happy to open a new issue.

davecheney:
@thsnr sorry I forgot to close the pull request. I am doing so now. 

Please lets continue this question on an issue. We don't do design and bug traige in pull requests. Thank you.

PR #168, opened 2018-08-30
Update .travis.yml

komuw:
update travisCI to also run builds using go1.11


AlekSi:
That one can be closed.

PR #175, opened 2018-11-30
make the output of embedded errors more readable

edwingeng:
sample code
``` go
err1 := errors.New("bomb 1")
err2 := errors.Wrap(err1, "bomb 2")
fmt.Printf("%+v\n", err2)
```
original output
```
bomb 1
main.main
	/Users/foo/golang/src/foo/x.go:20
runtime.main
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/proc.go:201
runtime.goexit
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/asm_amd64.s:1333
bomb 2
main.main
	/Users/foo/golang/src/foo/x.go:21
runtime.main
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/proc.go:201
runtime.goexit
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/asm_amd64.s:1333

```
new output
```
bomb 1
main.main
	/Users/foo/golang/src/foo/x.go:20
runtime.main
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/proc.go:201
runtime.goexit
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/asm_amd64.s:1333
====================
bomb 2
main.main
	/Users/foo/golang/src/foo/x.go:21
runtime.main
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/proc.go:201
runtime.goexit
	/usr/local/Cellar/go/1.11.2/libexec/src/runtime/asm_amd64.s:1333
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -245,7 +245,7 @@ func (w *withMessage) Format(s fmt.State, verb rune) {
 	switch verb {
 	case 'v':
 		if s.Flag('+') {
-			fmt.Fprintf(s, "%+v\n", w.Cause())
+			fmt.Fprintf(s, "%+v\n====================\n", w.Cause())
 			io.WriteString(s, w.msg)
 			return
 		}
--- a/format_test.go
+++ b/format_test.go
@@ -95,14 +95,14 @@ func TestFormatWrap(t *testing.T) {
 	}, {
 		Wrap(io.EOF, "error"),
 		"%+v",
-		"EOF\n" +
+		"EOF\n" + "=+\n" +
 			"error\n" +
 			"github.com/pkg/errors.TestFormatWrap\n" +
 			"\t.+/github.com/pkg/errors/format_test.go:96",
 	}, {
 		Wrap(Wrap(io.EOF, "error1"), "error2"),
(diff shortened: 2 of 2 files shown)


puellanivis:
One is not supposed to use `Wrap` when an error is known to already contain a stacktrace. And it is advised to avoid using `Wrap` unless you would otherwise use `New`.

Instead, unless you _know_ you need to add a stacktrace, use `WithMessage`, otherwise, you are creating a second stacktrace which in 90% of cases will be unnecessary, as it is almost bound to repeat a lot of the already contained stacktrace.

davecheney:
@edwingeng thanks for this PR. Sadly I'm not going to accept it because I have already rejected several others that change the output in various ways. I agree that your PR improves the output, but after this long people will be relying on the string output not changing.

I'm tidying up this project for a v1.0.0 release which will lock in the current behaviour. I'm not sure if there will be a v2 breaking change as people should probably move to the Go team's errors package when it becomes available.

PR #177, opened 2018-12-18
Add support for go-module

thedevsaddam:
Add support for go-module


davecheney:
Thanks but no thanks. This package has no dependencies, it does not need a go.mod file to be used by other packages.

PR #183, opened 2019-01-05
Switch to runtime.CallersFrames

### LINKED ISSUE #107: Migrate `runtime.Callers/2`
Given Slide 25 from [talk: Mid-stack inlining in the Go compiler (external)](https://docs.google.com/presentation/d/1Wcblp3jpfeKwA0Y4FOmj63PW52M_qmNqlQkNaLj0P5o/edit#slide=id.g1b2157b5d1_3_229) it seems like [`errors.callers`](https://github.com/pkg/errors/blob/a887431f7f6ef7687b556dbf718d9f351d4858a0/stack.go#L127) should migrate.

### LINKED ISSUE #160: Raising custom error types/ const errors
The io package declares an EOF error like this:

    var EOF = errors.New("EOF")

using this pattern to raise custom errors leads to ugly stack traces since the errors are generated during intialization, not when they happen:

````
github.com/andig/gosml.init
        /Users/andig/Documents/htdocs/go/src/github.com/andig/gosml/crc16.go:32
main.init
        <autogenerated>:1
runtime.main
        /usr/local/opt/go/libexec/src/runtime/proc.go:186
runtime.goexit
        /usr/local/opt/go/libexec/src/runtime/asm_amd64.s:2361
````

in fact, the entire call stack of where is error happens is lost which means using the error type is impossible.

How could one implement custom error types or errors the can be behaviour-checked according to your blog article?

davecheney:
Fixes #160
Fixes #107

Signed-off-by: Dave Cheney <dave@cheney.net>

### DIFF
--- a/stack.go
+++ b/stack.go
@@ -9,32 +9,26 @@ import (
 )
 
 // Frame represents a program counter inside a stack frame.
-type Frame uintptr
+type Frame runtime.Frame
 
 // pc returns the program counter for this frame;
 // multiple frames may have the same PC value.
-func (f Frame) pc() uintptr { return uintptr(f) - 1 }
+func (f Frame) pc() uintptr { return runtime.Frame(f).PC }
 
 // file returns the full path to the file that contains the
 // function for this Frame's pc.
 func (f Frame) file() string {
-	fn := runtime.FuncForPC(f.pc())
-	if fn == nil {
+	file := runtime.Frame(f).File
+	if file == "" {
 		return "unknown"
 	}
-	file, _ := fn.FileLine(f.pc())
 	return file
 }
 
 // line returns the line number of source code of the
 // function for this Frame's pc.
 func (f Frame) line() int {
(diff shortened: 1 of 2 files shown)


arcana261:
Hiiiiiiii

I wanted to provide a feed back guys! :)

I think since `errors.Frame` is exported type definition, it was not probably best to do this breaking change since probably lots of things depend on the fact that a `Frame` is a program counter. e.g. see this [issue on evalphobia/logrus_sentry](https://github.com/evalphobia/logrus_sentry/issues/74). I think if we wanted to provide a stable API and hide the fact of what `Frame` actually means, we had to wrap it inside a struct and make it un-exported :)

davecheney:
I’m sorry about this but this is unavoidable do to the change in the way mid stack inljning works. 

Please let’s continue this conversation on a _new issue_

davecheney:
@arcana261 see #188

PR #192, opened 2019-01-08
errors: detect unknown frames correctly

randall77:
With Go 1.12, we will now be doing mid-stack inlining. This exposes
inlined frames to runtime.Callers and friends.

The spec says that a runtime.Frame may have a nil Func field for
inlined functions. Instead, use the Function field to detect whether
we really have an empty Frame.

### DIFF
--- a/format_test.go
+++ b/format_test.go
@@ -360,6 +360,30 @@ func TestFormatGeneric(t *testing.T) {
 	}
 }
 
+func wrappedNew(message string) error { // This function will be mid-stack inlined in go 1.12+
+	return New(message)
+}
+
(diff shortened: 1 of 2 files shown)


randall77:
If you're going to go back to []uintptr for the stack, you should use the runtime.Callers to get the []uintptr, then subtract 1 before calling FuncForPC for each of the entries. That should still work with 1.12. (And don't subtract 1 from the result of runtime.Caller.)

Feel free to use the test, though. We have an internal Google test that was tripping up on the wrapper issue.

davecheney:
Thanks Keith, will do.

Issue #173, opened 2018-10-22
Cause - Last error with stack


marczahn:
Hi,

Let's assume we have some nested errors:

~~~
function sthBuiltInWhichReturnsAPlainError() error {
    return someBuiltErrorWithoutStack
}

func customFunctionWhichWrapsErrors() error {
    err := sthBuiltInWhichReturnsAPlainError()
    if err != nil {
        return errors.Wrap(err, "something happened")
    }
}

err := customFunctionWhichWrapsErrors()
errors.Cause(err) // <- Here I get the someBuiltErrorWithoutStack which has no stack
~~~

I would like to have the "last" error which has a stack. In this case it would be the one that is returned by customFunctionWhichWrapsErrors. But when I call errors.Cause I get the someBuiltErrorWithoutStack since this is the last one in the nesting.

Is this somehow possible without writing custom code?

dcormier:
> Is this somehow possible without writing custom code?

It isn't possible today. I just needed exactly that and ended up writing this:

```golang
func Stack(err error) errors.StackTrace {
	type causer interface {
		Cause() error
	}

	type stackTracer interface {
		StackTrace() errors.StackTrace
	}

	switch err := err.(type) {
	case stackTracer:
		return err.StackTrace()

	case causer:
		return Stack(err.Cause())
	}

	return nil
}
```

puellanivis:
This code example would return the first StackTrace. (It tests for `stackTracer` implementation before `causer`).

The original post wants to find the last StackTrace in the chain, rather than the first.

dcormier:
Yup. I wrote it to do what I needed. But it wouldn't take much to change it to return the stack from the last error that had one.

Issue #94, opened 2016-11-30
Proposal: add a package-level StackTrace() method


flimzy:
I think it would be convenient to have a package-level `StackTrace(err error) StackTrace` method, as a bit of an analog to the `Cause(err error) error` method, which returns the earliest stack trace found in the composite error value.

This is something I have implemented in at least three packages myself now, but it seems likely to be of value to others, so I'm beginning to think perhaps it belongs here.

The reason it's not sufficient to simply call `err.StackTrace()`, is that an error may have been `Wrap()`ed multiple times, in which case one generally (if not always) wants the _earliest_ stack trace.  And:

    cause := err.(causer)
    stackTrace := cause.(.stackTracer)

doesn't work, because generally the cause does not include a stack trace.

So my proposal is to add a package-level `StackTrace()` method such that one can do, for instance:

    if stackTrace := errors.StackTrace(err); stackTrace != nil {
        fmt.Printf("Stacktrace: %+v\n", stackTrace)
    }

If this proposal meets general approval, I can quickly submit a PR for further consideration.

davecheney:
I see the utility of this function, but like other similar requests for
wrapping errors if they haven't already been wrapped etc, or only wrapping
if haven't been annotated in a particular I'm wary because the logic is
subtle and the result of errors.StackTrace(err) would work for you, but I'm
sure someone else would want a single different version.

At the moment I'm punting on this by giving you to tools to write this
method yourself according to your specifications.

I know it's not the answer you were looking for, but it's thr state of
things today.

flimzy:
> I see the utility of this function, but like other similar requests for
wrapping errors if they haven't already been wrapped etc, or only wrapping
if haven't been annotated in a particular I'm wary because the logic is
subtle and the result of errors.StackTrace(err) would work for you, but I'm
sure someone else would want a single different version.

That's a fair criticism.  Thanks for considering.

davecheney:
I dobt want you see this as a criticism, just my caution that, looking at
the history of the issues raised in this repo, many people have different
requirements for how they want to report their errors, what detail, how and
where they want stack traces recorded, and how this is reported.

There is no one size fits all option, save providing only the bare building
blocks, and the only other option I see are to arbitrarily pick one way to
record and represent error, or provide a flotilla of helpers to attempt to
address all requirements. Neither of those sounds feasable.

flimzy:
[...]

PR #195, opened 2019-01-23
get first stack trace info.

wangsai-silence:
For this issue https://github.com/pkg/errors/issues/173

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -285,7 +285,6 @@ func Cause(err error) error {
 // the errors in the chain of Causes. If the error does not implement
 // Cause, the original error will be returned. If the error is nil,
 // nil will be returned without further investigation.
-
 func Stack(err error) StackTrace {
 	type causer interface {
 		Cause() error
--- a/errors.go
+++ b/errors.go
@@ -310,9 +310,5 @@ func Stack(err error) StackTrace {
 		err = causer.Cause()
 	}
 
-	if topStackInfo != nil {
-		return topStackInfo
-	}
-
-	return nil
+	return topStackInfo
 }
--- a/errors.go
+++ b/errors.go
@@ -280,3 +280,34 @@ func Cause(err error) error {
 	}
 	return err
 }
+
+// Stack returns the first stack trace of the errors, if possible.
+func Stack(err error) StackTrace {
+	type causer interface {
+		Cause() error
+	}
+
+	type stackTracer interface {
+		StackTrace() StackTrace
+	}
+
+	var stackErr error
+
+	for {
+		if _, ok := err.(stackTracer); ok {
+			stackErr = err
+		}
+
+		if causer, ok := err.(causer); ok {
+			err = causer.Cause()
+		} else {
+			break
+		}
+	}
+
+	if stackErr != nil {
+		return stackErr.(stackTracer).StackTrace()
+	}
+
+	return nil
+}
--- a/errors.go
+++ b/errors.go
@@ -281,7 +281,11 @@ func Cause(err error) error {
 	return err
 }
 
-// Stack returns the first stack trace of the errors, if possible.
+// Stack returns the earliest/deepest StackTrace attached to any of
+// the errors in the chain of Causes. If the error does not implement
+// Cause, the original error will be returned. If the error is nil,
+// nil will be returned without further investigation.
+
 func Stack(err error) StackTrace {
 	type causer interface {
 		Cause() error
--- a/errors.go
+++ b/errors.go
@@ -291,22 +291,24 @@ func Stack(err error) StackTrace {
 		StackTrace() StackTrace
 	}
 
-	var stackErr error
+	var topStackInfo StackTrace
 
 	for {
-		if _, ok := err.(stackTracer); ok {
-			stackErr = err
+		stackErr, ok := err.(stackTracer)
+		if ok {
+			topStackInfo = stackErr.StackTrace()
 		}
 
-		if causer, ok := err.(causer); ok {
-			err = causer.Cause()
-		} else {
+		causer, ok := err.(causer)
+		if !ok {
 			break
 		}
+
+		err = causer.Cause()
 	}
 
-	if stackErr != nil {
-		return stackErr.(stackTracer).StackTrace()
+	if topStackInfo != nil {
+		return topStackInfo
 	}
 
 	return nil
--- a/errors_test.go
+++ b/errors_test.go
@@ -106,9 +106,18 @@ func TestCause(t *testing.T) {
 	}
 }
 
+func printStack(tracer StackTrace) string {
+	var logInfo string
+	for _, f := range tracer {
+		logInfo += fmt.Sprintf("%+s:%d ", f, f)
+	}
+	return logInfo
+}
+
 func TestStack(t *testing.T) {
-	x := New("error")
-	y := WithStack(x).(*withStack).StackTrace()
+
+	x := WithStack(New("error"))
+	y := x.(*withStack).StackTrace()
 
 	tests := []struct {
 		err  error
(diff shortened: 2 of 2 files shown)


errors.go
```diff
@@ -280,3 +280,34 @@ func Cause(err error) error {
...
+	}
+
+	var stackErr error
+
+	for {
+		if _, ok := err.(stackTracer); ok {
+			stackErr = err
+		}
+
+		if causer, ok := err.(causer); ok {
```
puellanivis: Prefer:
```
		cause, ok := err.(causer)
		if !ok {
			break
		}
		err = cause.Cause()
```
The same as `Cause` in this package.

errors.go
```diff
@@ -280,3 +280,34 @@ func Cause(err error) error {
...
+func Stack(err error) StackTrace {
+	type causer interface {
+		Cause() error
+	}
+
+	type stackTracer interface {
+		StackTrace() StackTrace
+	}
+
+	var stackErr error
```
puellanivis: We could avoid having to re-type-assert the stackTracer if we recognize that `stackErr` is always going to be a `stackTracer` and enforce that through typing:
```
	var top stackTracker
	…
		if stack, ok := err.(stackTracer); ok {
			top = stack
		}
	…
	if top == nil {
		return nil
	}

	return top.StackTrace()
```

errors_test.go
```diff
@@ -106,6 +106,54 @@ func TestCause(t *testing.T) {
...
+	}
+
+	for i, tt := range tests {
+		got := Stack(tt.err)
+		if got == nil || tt.want == nil {
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("test %d: got %#v, want %#v", i+1, got, tt.want)
+			}
+
+			return
```
puellanivis:
```suggestion
			continue
```

Otherwise, we break out of the test after the first test, because definitively `tt.want == nil`.

Also, as written, by contract of the code design, we should expect all the other table tests to fail after this change.

wangsai-silence: You are right. I will rewrite it.

wangsai-silence: Thank you for your comments

errors.go
```diff
@@ -280,3 +280,36 @@ func Cause(err error) error {
 	}
 	return err
 }
+
+// Stack returns the first stack trace of the errors, if possible.
```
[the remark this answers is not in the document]
wangsai-silence: Thank you so much for your patience.  Your advice has been invaluable, and I have followed the suggest expect the last one. I don't know how to deal with this documentation. And I just wrote your words on it. So could you help to make it more clear?

errors.go
```diff
@@ -280,3 +280,40 @@ func Cause(err error) error {
 	}
 	return err
 }
+
+// Stack returns the earliest/deepest StackTrace attached to any of
+// the errors in the chain of Causes. If the error does not implement
+// Cause, the original error will be returned. If the error is nil,
+// nil will be returned without further investigation.
+
```
puellanivis: The docstrings should not have a space between them an the function. I don’t know for sure this would break godoc, but convention suggest we strictly enforce it.

wangsai-silence: OK. I'd remove it. Thank you.

errors.go
```diff
@@ -280,3 +280,39 @@ func Cause(err error) error {
 	}
 	return err
 }
+
+// Stack returns the earliest/deepest StackTrace attached to any of
+// the errors in the chain of Causes. If the error does not implement
+// Cause, the original error will be returned. If the error is nil,
```
flimzy: I think you mean "the stacktrace from the original error will be returned"?

errors.go
```diff
@@ -280,3 +280,39 @@ func Cause(err error) error {
...
+// nil will be returned without further investigation.
+func Stack(err error) StackTrace {
+	type causer interface {
+		Cause() error
+	}
+
+	type stackTracer interface {
+		StackTrace() StackTrace
+	}
```
flimzy: You shouldn't need to define these types here--these are already global types.

errors.go
```diff
@@ -280,3 +280,39 @@ func Cause(err error) error {
...
+		}
+
+		err = causer.Cause()
+	}
+
+	if topStackInfo != nil {
+		return topStackInfo
+	}
+
+	return nil
```
flimzy: You can collapse these 4 lines to simply:

    return topStackInfo

hanzei:
@wangsai-silence Could you please take a look at the comments above?

errors.go
wangsai-silence: Sorry, but I didn't find any public define of these types. Only private ones.

wangsai-silence:
> @wangsai-silence Could you please take a look at the comments above?

OK. I've done. Thank you for remind me about that.

errors.go
flimzy: This one is also private. Do you need a public one?

wangsai-silence: I'm not sure about that. I don't know why the previous ones are all private.

puellanivis: We want to keep the scope of the interfaces to the smallest set available. Firstly, because defining these interfaces scoped to the function that uses them makes it more clear what the interface is (it’s defined right above you).  Secondly, it also prevents us from exporting interface types unnecessarily (one cannot export something that is not even there) Finally, there’s considerations that we should be avoiding using interfaces as much as possible. Giving some overhead required to actually make use of them is a good way to deter rampant interface usage.

Basically, we should generally be avoiding defining package-global interfaces as much as possible.

mightyguava:
Oh I was just looking to implement this. @wangsai-silence @puellanivis what is left to be done for this PR?

wangsai-silence:
> Oh I was just looking for something like this. @wangsai-silence @puellanivis what is left to be done for this PR?

Nobody answered. I have used it in my project.

puellanivis:
🤷 I don’t have collaborator rights to the repo, so I cannot approve and merge. Also, I was going to leaving the approval/denial of the mere design up to other parties anyways.

The branch is however now out-of-date with the base branch, and likely would end up with a merge conflict or something.

Fortunately, the code works both inside of and outside of this package. So there is not much stopping anyone from replicating it. But, with proper `pkg/error` use sanitation, you can avoid the need to use it drastically.

flimzy:
FWIW, I [proposed something like this](https://github.com/pkg/errors/issues/94) back in 2016. While it wasn't outright rejected, I closed the issue because I agreed with Dave Cheney's cautious approach.

mightyguava:
@flimzy ah, asking each package to implement their own version makes sense with that link. I don't understand what `errors.As` has to do with this though. I want to print a stacktrace for debugging.

flimzy:
> @flimzy ah, asking each package to implement their own version makes sense with that link. I don't understand what `errors.As` has to do with this though. I want to print a stacktrace for debugging.

Two things:

1. I want a version that honors Go 1.13's `wrapper` interface. `errors.As` is the easiest way to address that.
2. `errors.As` makes it trivial to implement this functionality, making the need for supporting it in a library like this less urgent, IMO. Example (untested "whiteboard" code):

```golang
stackErr := new(interface{ StackTrace() errors.StackTrace })
for errors.As(err, stackErr) {
    err = stackErr
}
if stack != nil {
    stack := stackErr.StackTrace()
}
```

mightyguava:
`errors.As` doesn't unwrap unless necessary, so I think you have an infinite loop there.

flimzy:
> `errors.As` doesn't unwrap unless necessary, so I think you have an infinite loop there.

Possible. As I said, the code is back-of-napkin quality, untested. I don't think it affects the substance of what I was saying.

puellanivis:
You are correct, `errors.As` does not unwrap if the given `error` is already assignable to the second argument. That’s easy enough to fix with an `errors.Unwrap` call.

But I would additionally recommend against using `new()` for an interface. Because an interface is already reference semantics, it is almost universally an error to make a direct datatype that is a pointer to an interface. The only time a pointer to interface is not an error, is when it is a derived type, for instance when you have a subfunction that would like to assign into an interface value, rather than just use it, where it should be derefed at the subfunction call. So, these changes get us:
```
func Stack(err error) StackTrace {
	var stackErr interface{ StackTrace() errors.StackTrace }
	for errors.As(err, &stackErr) {
		err = errors.Unwrap(stackErr) // err will be assignable, and infinite loop otherwise.
	}
	if stackErr != nil {
		return stackErr.StackTrace()
	}
	return nil
}
```

flimzy:
> But I would additionally recommend against using `new()` for an interface. <snip>

All good advice. Please don't get hung up on the minutiae of a throw-away coding example.

puellanivis:
I don’t think we derailed things. We ended up with a corrected version, that people can use. It’s the good-ol’ “say a wrong answer, and you will immediately get 50 other people to do the work of finding the correct answer for you.”

PR #197, opened 2019-01-28
Add json.Marshaler support to the Frame type.

### LINKED ISSUE #196: Proposal: json.Marshaler support for stack traces
This may be a long shot, but I wonder if it would be worth while to add `json.Marshaler` support for stack traces, to make it easier to pass stack traces to, for instance, [logrus](https://github.com/sirupsen/logrus).

I'm thinking that implementing the `json.Marshaler` interface on the `errors.Frame` type might be the best approach. This would make it possible to log a stack trace as simply as:

    if stackErr, ok := err.(stackTracer); ok {
        log.WithField("stacktrace", stackErr.StackTrace())
    }

and the JSON output ought to be a nice array of stack trace frames.

This would of course raise the question of precisely _how_ to format these frames. I'm using my own implementation now which is essentially the `%+v` format verb's output, without newlines and tabs. Whatever format is chosen, consumers would be free to write their own marshaler and wrap the type, if necessary for special use cases.

Are there any drawbacks to this?  Other thoughts or considerations?

flimzy:
Fixes #196

This PR adds `json.Marshaler` support for the `Frame` type, to facilitate logging of stack traces in JSON format.  The output format is equivalent to `fmt.Printf("v+%")`, but without newlines or tab characters.

To me, this format seems the most generally useful, but an argument could be made for outputting a JSON object instead (i.e. `{"file":"...", "line": XX, "name": "..."}`.).

### DIFF
--- a/json_test.go
+++ b/json_test.go
@@ -1,0 +1,29 @@
+package errors
+
+import (
+	"encoding/json"
+	"regexp"
+	"testing"
+)
+
+func TestFrameMarshalJSON(t *testing.T) {
+	var tests = []struct {
+		Frame
+		want string
+	}{{
+		initpc,
+		`^"github.com/pkg/errors.init .+/github.com/pkg/errors/stack_test.go:\d+"$`,
+	}, {
+		0,
+		`^"unknown"$`,
+	}}
+	for i, tt := range tests {
+		got, err := json.Marshal(tt.Frame)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !regexp.MustCompile(tt.want).Match(got) {
+			t.Errorf("test %d: MarshalJSON:\n got %q\n want %q", i+1, string(got), tt.want)
+		}
+	}
+}
--- a/json_test.go
+++ b/json_test.go
@@ -6,6 +6,28 @@ import (
 	"testing"
 )
 
+func TestFrameMarshalText(t *testing.T) {
+	var tests = []struct {
+		Frame
+		want string
+	}{{
+		initpc,
+		`^github.com/pkg/errors\.init(\.ializers)? .+/github\.com/pkg/errors/stack_test.go:\d+$`,
+	}, {
+		0,
+		`^unknown$`,
+	}}
+	for i, tt := range tests {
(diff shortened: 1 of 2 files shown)


stack.go
```diff
@@ -83,6 +84,18 @@ func (f Frame) Format(s fmt.State, verb rune) {
 	}
 }
 
+// MarshalJSON formats a stacktrace Frame for easy inclusion in JSON stack
+// traces. The format is the same as that of fmt.Format("%+v"), but without
+// newlines or tabs.
+func (f Frame) MarshalJSON() ([]byte, error) {
```
puellanivis: Rather than just marshaling this to a string, then `json.Marshal`ing that, I think we would prefer `MarshalText` here.

From [the encoding/json references](https://golang.org/pkg/encoding/json/#Marshal) `If no MarshalJSON method is present but the value implements encoding.TextMarshaler instead, Marshal calls its MarshalText method and encodes the result as a JSON string.`

json_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+)
+
+func TestFrameMarshalJSON(t *testing.T) {
+	var tests = []struct {
+		Frame
+		want string
+	}{{
+		initpc,
+		`^"github.com/pkg/errors.init(.ializers)? .+/github.com/pkg/errors/stack_test.go:\d+"$`,
```
puellanivis: Please escape periods that must match periods.

flimzy: I just copied then adapted these REs from existing tests, but will do.

stack.go
flimzy: Good suggestion.

flimzy:
@puellanivis See updates. I left the json.Marshaler tests in tact, but maybe that can be removed as tacitly correct.

json_test.go
puellanivis: Indeed, it’s super common to miss, and rarely does it ever cause any problems.

Also, you missed one… :wink:

puellanivis:
Looks good from my end.

json_test.go
flimzy: Fixed that one now, too.

PR #205, opened 2019-05-30
add status in error fundamental

wNee:


davecheney:
Thank you for your PR however I am going to close it for the following reasons

1. This project operates on a talk then code policy, there must be a discussion of your proposed change in an issue before sending code.
2. Given the integration of an errors like package upstream, this project is effectively in maintenance mode. If you want to add a feature, you should do so upstream on golang/go

PR #212, opened 2019-11-01
travis.yml: add Go 1.13

aperezg:
Also deprecate Go 1.10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,5 @@ go:
   - 1.13.x
   - tip
 
-env:
-  - GO111MODULE=off
-
 script:
   - make check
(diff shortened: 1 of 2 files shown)


puellanivis:
The previous PR to change version information in travis is: [PR-200](https://github.com/pkg/errors/pull/200) which is simply titled, “travis.yml: add Go 1.12”

I don’t think it’s necessary to go into why 1.10 support is being dropped, it’s a natural part of upgrading to covering 1.13.

Next up, do things actually break such that we need `GO111MODULE=off` or is this simply a ”since there is no go.mod file, I’m assuming I have to turn off GO111MODULE”? If it does not actually break things to no specify it, then I see no good reason to change it now.

aperezg:
Thanks for your comments @puellanivis, they seem very reasonable to me so, I will do the changes that you're proposing

PR #213, opened 2019-11-03
feat: support std errors functions

Sherlock-Holo:
add function `Is`, `As` and `Unwrap`, like std errors, so that we can
continue to use pkg/errors with go1.13 compatibility


.gitignore
```diff
@@ -22,3 +22,5 @@ _testmain.go
 *.exe
 *.test
 *.prof
+
+.idea
```
aperezg: This is responsability of each developer remove in his own gitignore globally, for a library this is not part of his responsabillity

.travis.yml
```diff
@@ -4,6 +4,7 @@ go:
   - 1.10.x
   - 1.11.x
   - 1.12.x
+  - 1.13.x
```
aperezg: There another PR about travis.yml #212 and this is out of scope of your feature

errors.go
```diff
@@ -128,15 +128,15 @@ func (f *fundamental) Format(s fmt.State, verb rune) {
 	switch verb {
 	case 'v':
 		if s.Flag('+') {
-			io.WriteString(s, f.msg)
+			_, _ = io.WriteString(s, f.msg)
```
aperezg: I know that there are very site that recommend to ignore explicity the errors, but first of all is out of scope the feture in self and that can be a discussion of the way of the library do its standard. So, could you revert all this kind of change around the package?

errors.go
```diff
@@ -159,19 +159,21 @@ type withStack struct {
 
 func (w *withStack) Cause() error { return w.error }
 
+func (w *withStack) Unwrap() error { return w.error }
```
aperezg: The Unwrap method is already added on the PR #206 furthermore this PR will be merge soon so actually you wouldn't have to add it

go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/pkg/errors
```
aperezg: There are an opened PR #203 about go.mod and It has some comments, and It's seems that a complicated topic, maybe with the xerrors package would be necessary, we'll check it.

.gitignore
Sherlock-Holo: ok, I will delete this line

errors.go
Sherlock-Holo: I will revert change like this

errors.go
Sherlock-Holo: If I delete this, the unit test will be failed, however after merge PR #206 it will be success, so I keep the unit test and wait for PR #206

go.mod
Sherlock-Holo: If we don't want to add go.mod and don't use other dependencies, I can copy xerrors codes for <go1.13

Sherlock-Holo: I will wait for your option and decide to add go.mod or not

go113_test.go
```diff
@@ -0,0 +1,140 @@
...
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{
+			name: "with stack",
+			args: args{
+				err:    WithStack(err),
+				target: new(customErr),
```
puellanivis: So… there’s a lot of interesting features of `struct{}` types. In particular here, they can all point to the same address in memory. This means (per surprisingly), that all of these `target`s could be exactly the same value.

It would be better to mimic the [example from `errors.As`](https://golang.org/pkg/errors/#As) and use something like a `*os.PathError` or at least use something like a `type customError struct { msg string }` that will ensure that the type has a non-zero size. Then use a `*customError` which you then dereference for the `As(…)` call below.

You should then make sure that the `customError.msg` field is actually set to the expected value that you set inside your `baseError := &customError{ msg: "…" }` above.

Remember, the functionality of `As(…)` is not just that it returns true or false, but that it also assigns that matching value into the `target` parameter.

go113_test.go
```diff
@@ -0,0 +1,140 @@
+// +build go1.13
```
puellanivis: I don’t think there is any difference between your go1.13 test and your !go1.13 test, since it all depends upon which package it is wrapping: the standard `errors` package, or the `golang.org/x/xerrors` package.

go.mod
puellanivis: It would be nice if we could avoid the `xerrors` package, to avoid having any dependencies at all, but I don’t think that would be possible without copy-pasting in the relevant code from `golang.org/x/xerrors` which is also not a good idea.

But this `go.mod` would need to be appropriate for both `go1.13' builds as well as `go1.12` builds, which makes the issue even more difficult.

aperezg: For me the idea of including the xerrors package or other external dependency not sounds so well, but the alternative are maybe even worse... we could create our own implementation, but we need to check always the compatibility with standard library and check that the behaviour is the same... so it's a problem in a future.

I think that it's a great idea have these two methods, because the final user only will need to import one errors library, and for the migration to standard library could be more easy for them. But I don't know which could be the better solution.

Maybe @davecheney have another idea, or he think that these features are not necessary on this package.

puellanivis: You’re not wrong. Putting wrappers of these functions in this package allows it to continue being a drop-in replacement for the main `errors` package.

An interesting idea is that we could just not import the functions unless it’s go1.13 or higher. If you’re using these functions with a lower version, then the standard errors package won’t have these functions either. 🤷‍♀ 

I think we are in a situation, though, where there are no good answers.

aperezg: So you propose do a simple file like `errors_113.go`:

```go
// +build go1.13

package errors

import (
	"errors"
)

func Is(err, target error) bool {
	return errors.Is(err, target)
}

func As(err error, target interface{}) bool {
	return errors.As(err, target)
}
```

And its respective test of course, and only enable this features on the pkg/errors that it will compiles on go1.13, right?

So, I think that is a very interesting point, so if the users only continue using this library for stack trace, and the standard library get to implement it, they will can migrate it easily.  And for our part there not any breaking changes.

Sherlock-Holo: I don't think only add `Is` `As` `Unwrap` for >=go1.13 is a good idea. Those functions is a good helper to help us handle error, but <go1.13 users who use pkg/errors can't enjoy them. With those functions they can keep using pkg/errors and enjoy those functions

go113_test.go
Sherlock-Holo: I forgot this 

> Remember, the functionality of As(…) is not just that it returns true or false, but that it also assigns that matching value into the target parameter.

Let me add more check for this

go.mod
puellanivis: Users who use `errors` with <go1.13 still need to include `xerrors` to get the functions anyways.

aperezg: @Sherlock-Holo I think that philosophy of this library always has been add the lacks of the standard library have, and right now only is add the stack trace. So the users could right now, need that library allow a easily migration between pkg/errors and standard library, nothing more.

The actual users can continue using `xerrors` in previous versions or some of the forks of this package that including the `xerrors` inside for use it with previous version.

IMHO think that is a good idea add the support only to the >=go1.13 to the methods, `Is` and `As`

Sherlock-Holo: ok, let's only support `Is` `As` `Unwrap` for >=go1.13

jayschwa:
If this package is going into maintenance mode (according to the README), why add new API surface area? "Maintenance mode" and "drop-in replacement" seem like competing motivations, especially as the standard library beefs up its package.

abraithwaite:
> If this package is going into maintenance mode (according to the README), why add new API surface area? "Maintenance mode" and "drop-in replacement" seem like competing motivations, especially as the standard library beefs up its package.

The change is small, it improves the user experience, and probably improves interoperability of tooling as well.  I think it would be silly not to include it personally.

Sherlock-Holo:
@aperezg please see are there anything need to improve, if has, let me know to optimize codes :-)

go113_test.go
```diff
@@ -3,14 +3,154 @@
 package errors
```
puellanivis: This will need the `go1.13` build tag, since we will not be exporting these functions from `xerrors` for <go1.13.

Also, testing the functions in our own code is kind of silly anyways, because they are just a wrapper for the standard library functions. So, we’re actually testing code that is strictly not our domain.

Sherlock-Holo: actually [go113_test.go](https://github.com/Sherlock-Holo/errors/blob/support-std-errors/go113_test.go#L1) has build tag

puellanivis: :woman_facepalming:

go113.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	stderrors "errors"
+)
+
+func Is(err, target error) bool { return stderrors.Is(err, target) }
+
+func As(err error, target interface{}) bool { return stderrors.As(err, target) }
+
+func Unwrap(err error) error { return stderrors.Unwrap(err) }
```
aperezg: Maybe Unwrap could continue in normal package, because you can also use with a xerrors library? what do you think, @puellanivis, @Sherlock-Holo?

go113_test.go
aperezg: Yes, maybe the test around the `Is` and `As` only must check that the compatibility is in there.

I don't know if the best solution, but I think in something like that:

```go
func TestIs(t *testing.T) {
	sentinelError := stdlib_errors.New("sentinel error")
	wrap := Wrap(sentinelError, "wrap error")
	if !Is(wrap, sentinelError) {
		t.Errorf("Expected that '%v' error and the '%v' error should be of the same type", sentinelError, wrap)
	}
}

func TestAs(t *testing.T) {
	type myError struct {
		error
	}
	err := &myError{stdlib_errors.New("error")}
	wrap := Wrap(err, "wrap error")

	var tt *myError
	if !As(wrap, &tt) {
		t.Errorf("Expected that '%v' error and the '%v' error should be of the same type", err, wrap)
	}
}
```

go113.go
Sherlock-Holo: `Unwrap` doesn't use any black magic (errors use `reflectlite` not `reflect`), it just uses type assert, we could implement it too

go113_test.go
Sherlock-Holo: `TestIs` and `TestAs` not only check the compatibility, it also make sure `withStack` and `withMessage` can work with them.

errors.go
```diff
@@ -286,3 +286,13 @@ func Cause(err error) error {
 	}
 	return err
 }
+
+func Unwrap(err error) error {
```
jayschwa: Why reimplement this function rather than wrapping it? Someone using Go 1.12 or older will not be using the new style of errors. This creates a risk of incompatibility if Go 1.14 subtly changes its `Unwrap` implementation.

Sherlock-Holo: @aperezg emm.... this is a problem if go1.14 changes `Unwrap` implment, so... we should keep `Unwrap` >= go1.13?

aperezg: But why is the problem to maintain the Unwrap function like today? I understand that the Unwrap is an interface on a std library right? Sorry, I try to understand what is the point to change the implementation a wrapped with the standard library. I doubt that they implement a breaking change on that, no?

https://github.com/pkg/errors/blob/master/errors.go#L163

go113_test.go
puellanivis: By contract, `withStack` and `withMessage` work with `Is` and `As` by implementing `interface{ Unwrap(error) error }`. This is something that is not within our scope to test. We test that they implement `Unwrap` according to contract, everything else is on `Is` and `As` to test.

errors.go
puellanivis: go1.14 cannot break the semantics of `Unwrap` without violating the go1 compat guarantee.

> These same considerations apply to successive point releases. For instance, code that runs under Go 1.2 should be compatible with Go 1.2.1, Go 1.3, Go 1.4, etc., although not necessarily with Go 1.1 since it may use features added only in Go 1.2

puellanivis: This functionality should likely be folded into `Cause()` and `Unwrap()` here just calls `Cause()`.

c.f. https://github.com/pkg/errors/pull/215

jayschwa: > go1.14 cannot break the semantics of Unwrap without violating the go1 compat guarantee.

Right now, standard library's `errors.Unwrap` is only interested in an `Unwrap()` interface, and it will always have to do that to retain backwards compatibility. However, Go 1.14 could hypothetically update `errors.Unwrap` to *additionally* handle a new `Foobar()` interface. In that scenario, it would be painful for `pkg/errors` to roll its own implementation for a "drop-in replacement".

puellanivis: I think worrying too much into the future results in writing features and support for things that maybe never even are necessary.

We should provide a `Cause`/`Unwrap` in this library that supports _both_ this package’s wrapping, as well as the standard package wrapping. We should concern ourselves with _today_.

If go1.14 expands the functionality of `Wrap`, then we can deal with that change _then_.

It does not make sense to burn bridges before we even get to them.

jayschwa: Hypothetical edge cases aside, calling through to the standard library's `Unwrap` will support both flavors of wrapped errors, and it's simpler than reimplementing the function. I think the latest commit in this PR is doing the right thing in that regard.

Sherlock-Holo: in my option, >1.13 errors will be compatible with xerrors, and xerrors now still uses unwrap interface. if std errors unwrap change its implementation, it will break old xerrors user codes

Sherlock-Holo: and `withStack` `withMessage` implement the unwrap interface to support these functions

puellanivis: Right, `xerrors.Unwrap()` will support this package as well, but what about a package that only supports `causer`?

Sherlock-Holo: that is bad idea, adding these functions is trying to make sure we don't import two packages at the same time, it may confuse user when use pkg/errors when std errors

puellanivis: It does not matter what _we_ want, this “theoretical” is a third-party package that only implements `causer`, has been abandoned and no longer updated. So it will never implement “unwrapper”.

So, again. It does not matter what _we_ implement in this package, for `withMessage` and `withStack`. If some other third-party package implemented `causer` but not `unwrapper`… we still want to ourselves allow some way to extract the original error from both `causer`s and `unwrapper`s in any permutation.

jayschwa:
```go
func Unwrap(err) error {
    if unwrapped := errors.Unwrap(err); unwrapped != nil {
        return unwrapped
    }
    if causer, ok := err.(causer); ok {
        return causer.Cause()
    }
    return nil
}
```

Sherlock-Holo: @jayschwa that is great, if we unwrap an error from other lib and it uses old pkg errors, the unwrap can still return the original error

Sherlock-Holo: as [that](https://github.com/pkg/errors/pull/213#discussion_r345502202) says, 1.14 errors won't change implementation, so we could make unwrap continue in normal package @aperezg

puellanivis: That `Unwrap` only works for specific permutations of wrapping with the standard library and `causer`s.

In the wild, we should expect any possible permutation of wrapping, including multiple times being wrapped in an alternating manner.

Sherlock-Holo: `Unwrap` should only unwrap error one times, no matter the inner error implement unwrap interface or not, like std `Unwrap` do.

mfridman:
@jayschwa why return `nil` instead of the original `error` that was passed in? 

https://github.com/pkg/errors/pull/213#discussion_r346045555

I.e., `Unwrap` may not get a causer or unwrapper, so shouldn't it return the error without modification?

jayschwa:
> @jayschwa why return `nil` instead of the original `error` that was passed in?

Because we want to match the behavior of the [standard library's `Unwrap`](https://godoc.org/errors#Unwrap):

> Unwrap returns the result of calling the Unwrap method on err, if err's type contains an Unwrap method returning error. **Otherwise, Unwrap returns nil.**

errors.go
puellanivis: But this code would fail to fully unwrap: `Unwrap(fmt.Errorf("wrap1: %w", myPrivateCauserOnlyError(fmt.Errorf("wrap 3: %w", err), "wrap2:")))` where I would get the `fmt.Errorf("wrap3: %w", err)` error not the original `err`, which would not be the expected behavior.

puellanivis: 🤦‍♀ the standard library `Unwrap` does not continuously unwrap a function, until it reaches the end result, unlike how `Cause` in this package works.

johan-lejdung:
What's the status on this PR, it's been almost a month since the initiative to update this package started. I'd be happy to help out any way I can.

Sherlock-Holo:
you can tell us if you want these functions only support 113 or all version

Johan Lejdung <notifications@github.com> 于 2019年11月24日周日 20:46写道：

aperezg:
Sorry I was very busy this month, I'll try to catch up with all comments ASAP

Sherlock-Holo:
@aperezg I think this PR should focus on add this functions for 1.13, and if many<1.13 users want them, we cloud create another PR to help them

johan-lejdung:
I agree with that @Sherlock-Holo 

The pr should focus on adding support for the new `std errors` functionalities. Rebuilding those functionalities for use on <1.13 doesn't make as much sense and isn't as pressing of an issue

ajpetersons:
How soon could this get merged? I would really love to see this PR merged in master and tagged so we can use Go 1.13 functions.

Sherlock-Holo:
these days I am a little busy too…maybe this weekend I can update this pr
for just support these functions >=1.13

Artūrs Jānis Pētersons <notifications@github.com> 于 2019年12月6日周五 21:24写道：

aperezg:
For me, I think that this PR would be only for add support to the new features on Go 1.13, so I think that as is now seems good.

> these days I am a little busy too…maybe this weekend I can update this pr
for just support these functions >=1.13

What do yo need to update on this PR to be merged?

Sherlock-Holo:
@aperezg Nothing need to do before merge, now this PR only add new feature for >= 1.13

PR #215, opened 2019-11-12
Support Go 1.13 error chains in `Cause`

jayschwa:
Imagine module A imports module B and both use `pkg/errors`. A uses
`errors.Cause` to inspect wrapped errors returned from B. As-is, B
cannot migrate from `errors.Wrap` to `fmt.Errorf("%w", err)` because
that would break `errors.Cause` calls in A. With this change merged,
`errors.Cause` becomes forwards-compatible with Go 1.13 error chains.
Module B will be free to switch to `fmt.Errorf("%w", err)` and that will
not break module A (so long as the top-level project pulls in the newer
version of `pkg/errors`).

### DIFF
--- a/go113_test.go
+++ b/go113_test.go
@@ -9,7 +9,29 @@ import (
 	"testing"
 )
 
-func TestErrorChainCompat(t *testing.T) {
+func TestCauseErrorChainCompat(t *testing.T) {
+	err := stderrors.New("the cause!")
+
+	// Wrap error using the standard library
+	wrapped := fmt.Errorf("wrapped with stdlib: %w", err)
+	if Cause(wrapped) != err {
+		t.Errorf("Cause does not support Go 1.13 error chains")
+	}
+
+	// Wrap in another layer using pkg/errors
+	wrapped = WithMessage(wrapped, "wrapped with pkg/errors")
+	if Cause(wrapped) != err {
+		t.Errorf("Cause does not support Go 1.13 error chains")
+	}
+
+	// Wrap in another layer using the standard library
+	wrapped = fmt.Errorf("wrapped with stdlib: %w", wrapped)
+	if Cause(wrapped) != err {
+		t.Errorf("Cause does not support Go 1.13 error chains")
+	}
+}
+
+func TestWrapErrorChainCompat(t *testing.T) {
 	err := stderrors.New("error that gets wrapped")
 	wrapped := Wrap(err, "wrapped up")
 	if !stderrors.Is(wrapped, err) {
--- a/cause.go
+++ b/cause.go
@@ -1,0 +1,29 @@
+// +build !go1.13
+
+package errors
+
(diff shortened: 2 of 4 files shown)


cause_go113.go
```diff
@@ -0,0 +1,37 @@
...
+//
+// If the error is nil, nil will be returned without further
+// investigation.
+func Cause(err error) error {
+	type causer interface {
+		Cause() error
+	}
+
+	for err != nil {
+		if cause, ok := err.(causer); ok {
```
puellanivis: We should probably switch this to a type switch now:
```
switch v := err.(type) {
case causer:
	err = v.Cause()
case unwrapper:
	cause := errors.Unwrap(err)
	if cause == nil {
		return err
	}
	err = cause
default:
	return err
}
```

go113_test.go
```diff
@@ -4,10 +4,20 @@ package errors
 
 import (
 	stdlib_errors "errors"
+	"fmt"
 	"testing"
 )
 
-func TestErrorChainCompat(t *testing.T) {
+func TestCauseErrorChainCompat(t *testing.T) {
```
puellanivis: I would also recommend also testing a `WithMessage(fmt.Errorf("wrap2: %w", WithMessage(err, "wrap1"))), "wrap3")`

cause_go113.go
```diff
@@ -0,0 +1,37 @@
...
+// Cause returns the underlying cause of the error, if possible.
+// An error value has a cause if it implements the following
+// interface:
+//
+//     type causer interface {
+//            Cause() error
+//     }
+//
+// If the error does not implement the Cause interface, the Cause
+// function will fallback to the standard library's errors.Unwrap
```
puellanivis: Let’s be more explicitly clear about the contract we are building: will we return the original error of any arbitrary ordering of `Cause` and `Unwrap`? The text presented here is potentially ambiguous… hm, I think the original was somewhat ambiguous as well.

We should be clear that this function will keep working recursively until it finds an error that is both not a `causer` and would return `errors.Unwrap(err) == nil`.

jayschwa: If an `error` has both `Cause()` and `Unwrap()`, I think `Cause()` should take precedence since that matches the original behavior of this function.

puellanivis: That’s not what I am saying. If we have a `causer` wrapped by a `fmt.Errorf` wrapped error, which wraps a `causer`, what is the behavior?

The code as you have written would unwrap all layers (which I think it should), but we should document that it will recursively try unwrapping all layers until it gets an unwrapped error.

aperezg:
@jayschwa If I'm not wrong you're agree with the changes propose for @puellanivis so, could you do this changes, to try unblock this PR, thanks :)

jayschwa:
Pushed some changes:
- Rebased off master branch.
- Tried to refine function documentation based on feedback.
- Added alternating stdlib and `pkg/errors` layers to test case.

PR #217, opened 2020-01-03
remove unnecessary use of fmt.Sprintf

aperezg:


example_test.go
```diff
@@ -195,7 +195,7 @@ func Example_stackTrace() {
 func ExampleCause_printf() {
 	err := errors.Wrap(func() error {
 		return func() error {
-			return errors.Errorf("hello %s", fmt.Sprintf("world"))
+			return errors.Errorf("hello %s", "world")
```
ferhatelmas: Nice but then why don't we write directly `hello world` with `errors.New`?

PR #233, opened 2020-12-06
feat: add go.mod

kaka19ace:
Hi,  can i add go.mod file to the repo ？  

For the Ref:  https://golang.org/ref/mod#modules-overview
```
A module is identified by a module path, which is declared in a go.mod file...
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/pkg/errors
+
+go 1.13


puellanivis:
This has come up before regularly (please look at the closed PRs dealing with `go.mod`).

The stance has been and remains:
* this package has no dependencies, and thus does not need a file to track dependencies.
* downstream users of this package do not need this file in order to use this package.

As such, there is no actual need for this file.

joanlopez:
It's exactly what @puellanivis said. So, I close this issue. Thanks!

Issue #210, opened 2019-09-05
How to check if an error is wrapped or is original


xxjapp:
Need method to check if error is wrapped

davecheney:
If you had that method, what would that enable you to do?

xxjapp:
@davecheney 

If I had the method, I could differentiate wrapped errors with call stack information and original errors without call stack information. For original errors I could add some call stack information by myself.

xxjapp:
For wrapped errors, I think there is no need to wrap it again when wrapping error from other's source.

puellanivis:
In general, either a source should guarantee that all of its errors are wrapped, or none of them are.
If you absolutely must work with a broken design, then the following code does what you want it to do. Writing a helper function to do the same would save you no complexity:
```
if causer, ok := err.(interface{ Cause() error }); ok {
  // value is wrapped
}
```

xxjapp:
@puellanivis 

Thank you for your detailed answer. For some reasons, I have to work with such a broken design from legacy source.

puellanivis:
Tech Debt… we meet again old friend…

davecheney:
Thank you for this but I do not intend to merge it. I encourage you to fork the package to add additional methods. Thank you

PR #239, opened 2021-05-19
feat ExistStack: Add a public method to determine if the error has been appended to the stack.

wbflooksky:
non-intrusive:
Add a public method to determine if the error has been appended to the stack.
Avoid adding the stack repeatedly for errors.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -119,7 +119,7 @@ func ExistStack(err error) bool {
 		Cause() error
 	}
 	if value, ok := err.(causer); ok {
-		return ExistStack(err.Cause())
+		return ExistStack(value.Cause())
 	} else {
 		return false
 	}
--- a/errors.go
+++ b/errors.go
@@ -67,9 +67,10 @@
 //
 // New, Errorf, Wrap, and Wrapf record a stack trace at the point they are
 // invoked. This information can be retrieved with the following interface:
-type stackTracer interface {
-        StackTrace() errors.StackTrace
-}
+//
+//     type stackTracer interface {
+//             StackTrace() errors.StackTrace
+//     }
 //
 // The returned errors.StackTrace type is defined as
 //
--- a/errors.go
+++ b/errors.go
@@ -67,10 +67,9 @@
 //
 // New, Errorf, Wrap, and Wrapf record a stack trace at the point they are
 // invoked. This information can be retrieved with the following interface:
-//
-//     type stackTracer interface {
-//             StackTrace() errors.StackTrace
-//     }
+type stackTracer interface {
+        StackTrace() errors.StackTrace
+}
 //
 // The returned errors.StackTrace type is defined as
 //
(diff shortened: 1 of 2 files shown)


wbflooksky:
fix: https://github.com/pkg/errors/issues/210

wbflooksky:
fix: https://github.com/pkg/errors/issues/102

errors.go
```diff
@@ -106,6 +106,23 @@ func New(message string) error {
...
+	if _, ok := err.(stackTracer); ok {
+		return true
+	}
+	type causer interface {
+		Cause() error
+	}
+	if _, ok := err.(causer); !ok {
+		return false
+	}
+	return ExistStack(err.Cause())
```
puellanivis, line 124: Here, you attempt to call the receiver method `Cause` on a thing of type `error`, which does not expose that implementation.

errors.go
```diff
@@ -106,6 +106,23 @@ func New(message string) error {
...
+func ExistStack(err error) bool {
+	type stackTracer interface {
+		StackTrace() StackTrace
+	}
+	if _, ok := err.(stackTracer); ok {
+		return true
+	}
+	type causer interface {
+		Cause() error
+	}
```
puellanivis, line 120: Just checking `Cause() error` ignores the possibility of an `Unwrap() error` which is kind of the preferred way of unwrapping errors now. So, this would only work with errors from this package, and those designed specifically for pre-unwrap compatibility with this package, which is no longer a strong argument since go1.13.

errors.go
puellanivis, line 124: 🤔 Not sure if the Go compiler is smart enough to rework end-recursion into the more efficient for-loop.

wbflooksky:
@puellanivis already fix

errors.go
```diff
@@ -106,6 +106,24 @@ func New(message string) error {
...
+	if value, ok := err.(causer); ok {
+		return ExistStack(err.Cause())
+	} else {
+		return false
```
puellanivis, line 124: Every code path in the preceding `if … { … }` block returns from the function, therefore there is no reason to use an `else` block.

Considering using:
```
cause, ok := err.(causer)
if !ok {
  return false
}
return ExistStack(cause.Cause())
```

Please check the implementation of `Cause(err error)` below, and try and mirror that as much as possible.

errors_test.go
```diff
@@ -27,6 +27,23 @@ func TestNew(t *testing.T) {
 	}
 }
 
+func TestExistStack(t *testing.T) {
+	tests := []struct {
+		err  error
+		want bool
+	}{
+		{io.EOF, false},
+		{Wrap(io.EOF, "read error"), true},
```
puellanivis, line 36: Missing test cases: `WithStack`, `WithMessage`, and `errors.New`.

errors.go
wbflooksky: Thank you very much for your help

davecheney:
Thank you for this but I do not intend to merge it. I encourage you to fork the package to add additional methods. Thank you

PR #247, opened 2026-03-26
Reopen in maintenance mode

tonistiigi:
Maintainers from @moby project can take over keeping this repo available in maintenance mode.

From README:

> This package is in maintenance mode. New features that extend the scope of the package are not being accepted and should be implemented in other modules. Exceptions would be features that improve interoperability with the standard library's `errors` package when it receives new features, as `pkg/errors` should mostly be able to act as a drop-in replacement. Bug fixes and reports are welcome. CI will be maintained to make sure the package is tested against new versions of Go and Go linting tools.

@thajeztah


crazy-max (commented):
Needs to bump actions to latest stable following Node 20 EOL coming soon on GitHub Hosted runners

README.md
```diff
@@ -1,4 +1,4 @@
-# errors [![Travis-CI](https://travis-ci.org/pkg/errors.svg)](https://travis-ci.org/pkg/errors) [![AppVeyor](https://ci.appveyor.com/api/projects/status/b98mptawhudj53ep/branch/master?svg=true)](https://ci.appveyor.com/project/davecheney/errors/branch/master) [![GoDoc](https://godoc.org/github.com/pkg/errors?status.svg)](http://godoc.org/github.com/pkg/errors) [![Report card](https://goreportcard.com/badge/github.com/pkg/errors)](https://goreportcard.com/report/github.com/pkg/errors) [![Sourcegraph](https://sourcegraph.com/github.com/pkg/errors/-/badge.svg)](https://sourcegraph.com/github.com/pkg/errors?badge)
+# errors [![GoDoc](https://godoc.org/github.com/pkg/errors?status.svg)](http://godoc.org/github.com/pkg/errors) [![Report card](https://goreportcard.com/badge/github.com/pkg/errors)](https://goreportcard.com/report/github.com/pkg/errors) [![Sourcegraph](https://sourcegraph.com/github.com/pkg/errors/-/badge.svg)](https://sourcegraph.com/github.com/pkg/errors?badge)
```
crazy-max: Go docs lives on `pkg.go.dev`

```suggestion
# errors [![GoDoc](https://godoc.org/github.com/pkg/errors?status.svg)](https://pkg.go.dev/github.com/pkg/errors) [![Report card](https://goreportcard.com/badge/github.com/pkg/errors)](https://goreportcard.com/report/github.com/pkg/errors) [![Sourcegraph](https://sourcegraph.com/github.com/pkg/errors/-/badge.svg)](https://sourcegraph.com/github.com/pkg/errors?badge)
```

crazy-max (approved):

bench_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build go1.7
 // +build go1.7
```
thaJeztah: With the go-module having `go1.20` as minimum, we can remove this build-tag

go113.go
```diff
@@ -1,3 +1,4 @@
+//go:build go1.13
 // +build go1.13
```
thaJeztah: Same here; probably can drop this, as the module requires go1.20

And maybe rename the file; we can merge it into the other one I guess?

go113_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build go1.13
 // +build go1.13
```
thaJeztah: Same here

json_test.go
```diff
@@ -12,13 +12,13 @@ func TestFrameMarshalText(t *testing.T) {
...
 		initpc,
-		`^github.com/pkg/errors\.init(\.ializers)? .+/github\.com/pkg/errors/stack_test.go:\d+$`,
+		`^github.com/pkg/errors\.init(\.ializers)? .+/stack_test.go:\d+$`,
 	}, {
 		0,
 		`^unknown$`,
 	}}
 	for i, tt := range tests {
-		got, err := tt.Frame.MarshalText()
+		got, err := tt.MarshalText()
```
thaJeztah: Curious; were these because of the [`QF1008`](https://staticcheck.dev/docs/checks/#QF1008)   (Omit embedded fields from selector expression) check? I tend to disable that one as it's (IMO) too opinionated, and in some cases it's more transparent to be explicit, and _not_ depend on embedding.

format_test.go
```diff
@@ -27,7 +27,7 @@ func TestFormatNew(t *testing.T) {
 		"%+v",
 		"error\n" +
 			"github.com/pkg/errors.TestFormatNew\n" +
-			"\t.+/github.com/pkg/errors/format_test.go:26",
+			"\t.+/format_test.go:26",
```
thaJeztah, line 30: Curious about these; were the tests wrong, or did behavior change between Go versions? (ISTR something changed in some of stdlib).

If it changed in Go, was that before, or after go1.20 (otherwise tests won't be valid with the minimum version specified in go.mod)

unwrap.go
```diff
@@ -1,6 +1,3 @@
-//go:build go1.13
-// +build go1.13
-
 package errors
```
thaJeztah, line 1: I would squash the removal of the tags with the previous commits, but keep the removal of the Makefile in this one.

astype.go
```diff
@@ -0,0 +1,10 @@
+//go:build !go1.26
+
+package errors
```
thaJeztah, line 3: This commit starts the work of picking up maintenance; I think it would make sense to move this (and forward) separate.

(Steps before this were just setting the baseline)

bench_test.go
```diff
@@ -23,7 +23,7 @@ func yesErrors(at, depth int) error {
 
 // GlobalE is an exported global to store the result of benchmark results,
 // preventing the compiler from optimising the benchmark functions away.
-var GlobalE interface{}
+var GlobalE any
```
thaJeztah, line 26: `any` was added in go1.18, and the module already had `go1.20` as minimum; I think we can squash this one with the "fix linting" commit, or move it before the addition of the go1.26 features.

.github/workflows/ci.yml
```diff
@@ -1,14 +1,62 @@
...
+  contents: read
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
-  workflow_dispatch:
   push:
+    branches:
+      - master
```
thaJeztah, line 12: Perhaps add `main` to the list, so that we can rename the default branch.

This commit could also be put before the go1.26 feature commit, making it part of the "baseline".

thaJeztah:
Left some comments (sorry I reviewed per commit, so some got superseded in later commits).

Overall looks good; some things I think we should do;

- Some commits can be squashed (see my comments)
- Move the fix linting, remove old CI files, and general cleanups first
- Move the go1.26 enhancement as a follow-up pull-request

format_test.go
tonistiigi: These tests were pre-gomod and expected checkout to be in GOPATH that is not a requirement anymore.

puellanivis:
I’m not sure the stacktrace feature is even useful. Most people tend to declare their errors at the top level, where the stacktrace ends up being `_init` anyways, and not anything useful pointing to where the error originated…

unwrap.go
tonistiigi: These are opinionated changes, while the first commit is purely making the linter and test pass with the recent version of Go without changing anything about the code. Moved the makefile away.

astype.go
tonistiigi: Moved to last

bench_test.go
tonistiigi: modernize is not same as lint. These changes make the code use the latest conventions but are not required for the code to work properly and lint pass in recent Go.

.github/workflows/ci.yml
tonistiigi: Renaming branches does not seem compliant with maintenance mode.

tonistiigi:
@puellanivis There is no obligation to use this package. This package works well with typed/static/chained errors; you add the stack trace to the return path, not to the definition path, if you use static errors.

errors.go
```diff
@@ -109,7 +109,7 @@ func New(message string) error {
 // Errorf formats according to a format specifier and returns the string
 // as a value that satisfies error.
 // Errorf also records the stack trace at the point it was called.
-func Errorf(format string, args ...interface{}) error {
+func Errorf(format string, args ...any) error {
```
puellanivis, line 112: 🤔 With the now standard library `Errorf` using the `%w` verb, this function is now subtly different from how likely a large number of people are people are using `fmt.Errorf`. This creates a situation where we’re likely to create significant surprise.

unwrap_test.go
```diff
@@ -170,9 +168,72 @@ func TestUnwrap(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
-			if err := Unwrap(tt.args.err); !reflect.DeepEqual(err, tt.want) {
+			if err := Unwrap(tt.args.err); err != tt.want {
 				t.Errorf("Unwrap() error = %v, want %v", err, tt.want)
 			}
 		})
 	}
 }
+
+func TestJoin(t *testing.T) {
```
puellanivis, line 178: Since our `Join` is just `return stderrors.Join(errs...)`, testing this is perhaps not particularly appropriate? It is, after all, not really our responsibility; it’s the standard library that has written the tests.

unwrap_test.go
```diff
@@ -170,9 +168,72 @@ func TestUnwrap(t *testing.T) {
...
+func TestJoinNil(t *testing.T) {
+	if err := Join(); err != nil {
+		t.Errorf("Join() = %v, want nil", err)
+	}
+	if err := Join(nil, nil); err != nil {
+		t.Errorf("Join(nil, nil) = %v, want nil", err)
+	}
+}
+
+func TestAsType(t *testing.T) {
```
puellanivis: The test here is really to more to ensure that `Wrap` errors properly process through `AsType`, rather than testing just `AtType` alone… which like `TestJoin` above, would have questionable value.

That is, the name of the function should probably reflect that we’re testing roundtripping `Wrap` errors, not just `AsType` all on its own.

unwrap_test.go
```diff
@@ -170,9 +168,72 @@ func TestUnwrap(t *testing.T) {
...
+	got, ok := AsType[customErr](wrapped)
+	if !ok {
+		t.Fatal("AsType[customErr]() = false, want true")
+	}
+	if got != err {
+		t.Errorf("AsType[customErr]() = %v, want %v", got, err)
+	}
+}
+
+func TestAsTypeNotFound(t *testing.T) {
```
puellanivis, line 245: Unlike `TestAsType` above, and more like the `TestJoin`, we’re basically testing either `stderrors.As` or `stderrors.AsType`… facing the same question of whose responsibility that testing belongs with, us, or `stderrors`.

puellanivis (commented):
Yes, that was the going advice to people at the time: that global errors should use `stderrors` since this package’s stacktraces are really only useful when they appear at the return.

My point of raising this point was that use of this package has always been with significant special caveats, and footguns. It was just that the calculus worked out in favor of using it, since one could wrap and unwrap errors.

Now, the whole utility of this package is really just the `WithStack` function, and I’m not really sure that the same calculus holds anymore. It might be better to just propose and implement a feature request on the standard `errors` package and add stack tracing to that.

It’s basically: there’s a reason why this package fell into an unmaintained state, and it was not really for lack of care or interest to maintain it.

errors.go
tonistiigi: `pkg/errors` is supposed to be drop-in replacement for `stderrors` where possible, but not replacement for `fmt`. So I don't see any issue with them being different.

unwrap_test.go
tonistiigi: This protects the cases where 1) somebody changes implementation in pkg/errors 2) implementation of `Join` is updated in a new version of Go. If 2 should happen, then it is detected in CI when updating Go, and maintainers can make a decision on how they want to handle the update.

tonistiigi:
> It might be better to just propose and implement a feature request on the standard errors package and add stack tracing to that.

Afaik this has been attempted multiple times with no progress

errors.go
puellanivis: Yes, I understand and I see that… but people are _going_ to expect that this `errors.Errorf` is compatible with `fmt.Errorf`.

Before `fmt.Errorf` existed, this was fine, because it didn’t have any mistaken conflations possible, it was kind of a standalone “hey, here’s a convenient `errors.New(fmt.Sprintf(format, args...))` for you,” but now if we’re reviving this module for a modern audience, they’re going to come with an expectation of compatibility between `fmt.Errorf` and `pkg/errors.Errorf`. Especially when the function signatures are identical.

unwrap_test.go
puellanivis: As you say above, this is supposed to be a drop-in replacement for `stderrors`. If we’re just aliasing `stderrors.Join`, then we’re not implementing anything that needs testing. We automatically comply with every semantic and behavior that `stderrors.Join` has.

There’s just zero code that we’re responsible for testing. And case 2 _is not guaranteed to break_ if `stderrors.Join` changes. It may change in a way that is outside of the scope of our tests, and that’s almost guaranteed, because go1 compat guarantee kind of suggests that the only way defined behavior will change is if there was a specification bug, security issue, or other unavoidable change of behavior.

errors.go
tonistiigi: Extending the scope is not compliant with the maintenance mode definition in the readme. `pkg/errors` never supported `%w`, so nobody can expect it to have the same implementation as `fmt`, and it has a different method for wrapping errors with extra context. If you prefer `%w` then continue using `fmt`, if you prefer `errors.Wrap`, then use `pkg/errors`.

unwrap_test.go
tonistiigi: Any public function exposed by this library should have regression testing. Doesn't matter if `Join` atm has handwritten implementation or calls into another library internally. Behavior can change (subtly) either in this repo or in upstream. If we declare that we will never make bugs and upstream will never make any changes, there is no need to test anything. 

> this is supposed to be

That's exactly the case. It doesn't say that `pkg/errors` is a direct alias or always the same as `stdlib`. If something breaks, then maintainers can decide if `pkg/errors` should adopt new stdlib behavior like the guidance says or keep backward compatibility.

puellanivis: Except that if I upgrade to `go1.XY` and `stderrors.Join` changes behavior, and `pkg/errors.Join` is (as is now) just a wrapping of that function, then my upstream binary will use that simple wrap of `stderrors.Join` and get the behavior of the new `stderrors.Join`, even if later builds of `pkg/errors` fails to compile because of the change in behavior.

That’s how to the go module dependency system works. `pkg/errors` gets whatever the `stderrors` package or other dependency is, regardless of what it itself declares in the `go.mod`.

This is exactly how the MySQL people got burned by the `url.Parse` breaking behavior change: https://github.com/golang-migrate/migrate/issues/264

Either you need to commit to maintaining your own implementation, or you accept that you’re wrapping `stderrors.Join` and getting whatever behavior it decides you’re going to have, even if your own tests break, my module is not running your tests.

errors.go
puellanivis: I’m saying that as a person who has been here providing code reviews and support since the module was active, and been “here” this whole time, able to give support: this is footgun.

If we had bothered to extend support for `pkg/errors` into the `fmt.Errorf` days, we would have definitely changed it to support `%w`, because that’s what users are going to expect.

kaovilai:
Just redirect to https://github.com/cockroachdb/errors already.

PR #250, opened 2026-07-21
fix: omit leading newline when formatting StackTrace with %+v

### LINKED ISSUE #230: StackTrace.Format with "%+v" includes a leading newline
https://play.golang.org/p/9pIaAGoKF1i

Fixing this might break formatting elsewhere, so needs to be done carefully.

Solaris-star:
## Summary
`fmt.Sprintf("%+v", stackTrace)` prefixed every frame with a newline, including the first one. That made a bare StackTrace print with a leading blank line (#230).

## Fix
In `StackTrace.Format` for `%+v`, write a newline only between frames (not before the first). Error formatting is unchanged because `stack.Format` still prefixes each frame with `\n` after the message.

## Test plan
- [x] `go test -vet=off .`
- [x] Updated `TestStackTraceFormat` expectation
- [x] New `TestStackTraceFormatNoLeadingNewline`

Fixes #230


puellanivis:
This is intended behavior.

Solaris-star:
Closing this based on maintainer feedback that the leading newline is intended behavior. Thanks for taking a look.

